PR #5, opened 2021-01-31
chore(Dropdowns): Migrate #97 - Refactor functionality and state management

andrewzulaybar:
## Description

The way that I set up state management for promotions currently was a bit non-conventional and difficult to maintain. So I thought I should change it to be a bit more like [Redux](https://redux.js.org/tutorials/fundamentals/part-3-state-actions-reducers).

Here's the simple breakdown of how it works and why/when this approach works:
Screen Shot 2020-12-31 at 9 15 28 PM
Screen Shot 2020-12-31 at 9 15 52 PM

So in our app, we can make use of this pattern to manage which promotions are currently being shown and in what order (filtering and sorting) according to which options have been selected for each dropdown. We have three main parts:
- The context provider component in `contexts/PromotionsList.tsx`: 
  - `PromotionsListProvider` component is what is used in the component hierarchy, allowing any of it's child components to make use of the context object (which we can access using the `usePromotionsList` function
- The reducer function in `reducers/PromotionsList.ts`
  - This function is responsible for taking the previous state object, along with the parameters to the dispatch call and constructing the new state. Basically, this function is called whenever you see a call to `dispatch`.
- The components making use of the context: 
  - `DropdownMenu.tsx`: Dropdown menu on the home page, used here because it allows us to clear all filters
  - `PromotionList.tsx`: List of promotions on home page, used here because we need to filter and sort the promotions whenever the state has been updated
  - `Home.tsx`: Home screen, used here because this is where we define the actions for the dropdowns

I've tried to add documentation wherever I could to make it easier to understand, but let me know if I should add some more in certain places. Let me know if there's anything that needs further explanation!


src/utils/routes.ts
```diff
@@ -0,0 +1,9 @@
+export default {
+  ENUMS: {
+    CUISINE_TYPES: '/enums/CuisineType',
+    DAYS_OF_WEEK: '/enums/Day',
+    DISCOUNT_TYPES: '/enums/DiscountType',
+    PROMOTION_TYPES: 'enums/PromotionType',
+  },
+  PROMOTIONS: '/promotions',
+};
```
ericrkuo: Awesome I like this a lot, will make it much easier to know where all the routes are at. 

What do you think about adding in the rest of the routes, for e.g.
```
USERS: '/users',
DISCOUNTS: '/discounts', <-- don't think we'll be using this much
SAVED_PROMOTIONS: 'saved_promotions`,
SCHEDULES: 'schedules` <-- this one too
```

src/services/PromotionService.ts
```diff
@@ -0,0 +1,113 @@
...
+
+type PromotionsResponse = PromotionsSuccess | PromotionsError;
+
+/**
+ * Fetches entire list of promotions. If a query object is given, filters the promotions according to the given query.
+ * If an error occurs, an empty list will be returned.
+ *
+ * @param query [optional] - An array of objects with key-value pairs for the query parameters
+ */
+export async function getPromotions(query?: Record<string, string>[]): Promise<Promotion[]> {
```
ericrkuo: Thanks for the documentation, this helps a lot

ericrkuo: Just out of curiosity, what's the reason behind using functions rather than classes?

src/screens/Home.tsx
```diff
@@ -3,152 +3,153 @@ import React, { ReactElement, useEffect, useState } from 'react';
...
+      {
+        endpoint: Routes.ENUMS.PROMOTION_TYPES,
+        setOptions: setPromotionTypes,
+      },
+    ];
+    endpoints.forEach(({ endpoint, setOptions }) => {
+      getEnum(endpoint)
+        .then((options) => setOptions(options))
+        .catch(() => setOptions([]));
+    });
```
ericrkuo, line 54: Would these endpoints be called each time we refresh the page? Wondering if it would be more efficient if you just have an `/enums` endpoint you could call once, which would return all the enums and their values. 

```
{
  PROMOTION_TYPES: [...],
  DISCOUNT_TYPES: [...],
  etc.
}
```

src/services/PromotionService.ts
ericrkuo: ALso is there some way we can make this typed rather than a `Record`? For example an interface like `PromotionQuery`? 

E.g. in BE we have this DTO for queries
https://github.com/ubclaunchpad/PromoPal-backend/blob/3c50736cd1a8065bdbf8711b05c990e2f18f93ad/src/main/validation/PromotionQueryValidation.ts#L39

src/utils/routes.ts
andrewzulaybar: Yeah good idea, I can add these!

src/services/PromotionService.ts
andrewzulaybar: @ericrkuo Good question, I think initially I tried using classes but they were hard to manage. Either I would have to instantiate it then pass that instance down the component hierarchy or I could make it a singleton instance (and not sure if I liked that). But this was before I added state management so now it might be easier to do since I can instantiate a Promotions object and then store that instance in the state object. I think I'll actually explore this idea, thanks for bringing it up :)

src/screens/Home.tsx
andrewzulaybar: Ohh like an `/enums` endpoint on the BE? Actually, this might be helpful because you're right these would be called every time we refresh the page unless we persisted these to local storage -- and if we do that we'd have to implement some periodic invalidation of the local storage... What do you think?

src/services/PromotionService.ts
andrewzulaybar: Ahh good point, yeah I can change this!

src/screens/Home.tsx
ericrkuo: Yeah so that way if we add more enums, you don't need to make as many changes here and it can all be in one http request. Not sure if that's common practice, but I think it would be beneficial. Yeah I think we can do an `/enums` endpoint in the BE, and then potentailly worry about local storage afterwards

src/services/PromotionService.ts
ericrkuo: Ohhh I see, that makes sense!

ericrkuo (approved):

PR #7, opened 2021-01-31
chore(Promotions): Migrate #85 - Add styling for promotion card

andrewzulaybar:
## Description

This change just adds styling to the promotion cards to implement the hi-fi designs. Currently, I just hardcoded the image URL and schedule for each promotion so they're all showing up as the same but this should be updated to use the data from the backend.

## Photos
Screen Shot 2021-01-30 at 11 21 27 PM

### DIFF
--- a/src/components/promotion/PromotionDetails.tsx
+++ b/src/components/promotion/PromotionDetails.tsx
@@ -4,7 +4,7 @@ import { ClockCircleOutlined, HeartOutlined } from '@ant-design/icons';
 import { Col, Row, Typography } from 'antd';
 import React, { CSSProperties, ReactElement } from 'react';
 
-import { Promotion } from '../../types/promotion';
+import { Promotion, Schedule } from '../../types/promotion';
 
 const { Title, Text } = Typography;
 
@@ -53,6 +53,7 @@ export default function PromotionDetails({
   description,
   expirationDate,
   restaurantName,
+  schedules,
 }: Promotion): ReactElement {
   /**
    * Returns display text for age of promotion.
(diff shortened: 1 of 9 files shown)


src/components/promotion/PromotionDetails.tsx
```diff
@@ -1,42 +1,181 @@
...
+  /**
+   * Returns display text for age of promotion.
+   *
+   * - If dateAdded < 1 day, displays "X hours ago"
+   * - If 1 day <= dateAdded < 1 week, displays "X days ago"
+   * - If 1 week <= dateAdded < 1 month, displays "X weeks ago"
+   * - If 1 month <= dateAdded < 1 year, displays "X months ago"
+   * - If 1 year <= dateAdded, displays "X years ago"
+   */
+  const promotionAge = (dateAdded: string) => {
```
ericrkuo: I wonder if we can leverage an NPM package for this, maybe something like https://www.npmjs.com/package/time-ago? 

Regardless, this is really cool

src/components/promotion/PromotionImage.tsx
```diff
@@ -1,8 +1,22 @@
...
+  image: {
+    borderRadius: 15,
+    minHeight: '100%',
+    objectFit: 'cover',
+    width: 100,
+  },
+};
+
+export default function PromotionImage({
+  // TODO: remove default image
```
ericrkuo: Maybe we can link the issue to this comment as well? https://github.com/ubclaunchpad/PromoPal-backend/issues/101?

ericrkuo (approved):

src/components/promotion/PromotionDetails.tsx
andrewzulaybar: Oooh good idea, what do you think of https://date-fns.org/? It seems like it has a lot of support and we could potentially use their other functions if we have needs for it in the future

Usage ex:
```
formatDistance(subDays(new Date(), 3), new Date())
```

src/components/promotion/PromotionImage.tsx
andrewzulaybar: Ahhh yes I'll add this thanks!

src/components/promotion/PromotionDetails.tsx
ericrkuo: Oooh that looks even better, let's go w that then!

src/components/promotion/PromotionDetails.tsx
```diff
@@ -1,42 +1,139 @@
...
+
+      const hourNum = parseInt(hour);
+      const minuteNum = parseInt(minute);
+
+      let formattedTime = `${hourNum === 12 ? hourNum : hourNum % 12}`;
+      if (minuteNum !== 0) {
+        formattedTime += `:${minuteNum}`;
+      }
+      formattedTime += hourNum < 12 ? ' AM' : ' PM';
+      return formattedTime;
```
ericrkuo, line 78: Does `date-fns` have something similar to this operation?

PR #12, opened 2021-02-01
chore: Add commitlint hook

andrewzulaybar:
## Description

Adds a hook on every commit that checks whether the commit message adheres to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/). This is mainly just to keep commits coherent and could make it useful in the future if we ever decide to release changelogs.

Here's a quick little guide with examples that I used to get started:
https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13


ericrkuo (approved):
This is really cool!

PR #15, opened 2021-02-06
PP-66: Add search query functionality

ericrkuo:
Leveraging BE FTS (Full-Text Search), we can now have bold descriptions and bold names for promotions
- E.g. the search keyword is "fry", FTS is smart enough to know "fries" is similar.
- results are ordered by rank (rank deems which promotions are most relevant to the given keyword)
image

@andrewzulaybar Also I noticed when we refresh, the filters are not remembered nor the search query. I think typically in applications, things like this are rememebred because they are in the url. For example, if we had cuisine "japanese" selected, the url would have something like `localhost:3000?cuisine="japansese"`. THat way when we refresh, we use the url to remember what filters were selected. That would be realy nice if we could support this, but probably not until the future.

Edit* Thanks @andrewzulaybar for helping out. I'll change the base branch to `main` once the `filter-promotions` PR gets merged in


andrewzulaybar:
@ericrkuo Thanks for getting started on this!

I took a look and added a few things in response to the things you mentioned (putting them down here for future reference):
- I pulled the `PromotionsListProvider` component higher up in the component hierarchy (placed it in the App.tsx file) so that the `NavigationBar` component is now a descendant
- I think you're right, we should probably go with users only being able to either use filters or perform a search as individual actions instead of having to worry about interactions between the two. I added some logic to reset the dropdowns when the search bar is used, but not the other way around (search text remains when dropdowns are used after searching the first time)
- With those changes, styling should be ok now! The dropdown does do that when the screen is small because we haven't implemented any responsiveness yet :(

Also that's a good suggestion, thanks for catching that! I'll add a ticket to add that in for the next sprint. Besides that, everything looks good! I'll just add some minor comments on this PR to clean things up a bit

src/components/PromotionList.tsx
```diff
@@ -43,7 +44,19 @@ export default function PromotionList({
...
+   * When a search query is set, fetches promotions that satisfy the query.
+   */
+  useEffect(() => {
+    if (state.searchQuery) {
+      getPromotions([{ searchQuery: state.searchQuery }]).then((promotions) => {
+        dispatch({ type: DispatchAction.DATA_SUCCESS });
+        setPromotions(promotions);
+      });
+    }
+  }, [dispatch, state.searchQuery]);
```
andrewzulaybar: Pulled this into a separate useEffect, because after playing around with it a bit I realized that we'll probably want this code block to run only when the searchQuery changes. Previously, if we first used the search bar and then tried to use the dropdowns, it wasn't working because `state.SearchQuery` was defined (so it was always going into the if-then case).

src/components/promotion/PromotionDetails.tsx
```diff
@@ -24,15 +23,17 @@ export default function PromotionDetails({
   description,
   expirationDate,
   restaurantName = 'Restaurant',
+  boldName = undefined,
+  boldDescription = undefined,
```
andrewzulaybar: I think you can remove the `= undefined` part because it should be undefined by default if that property isn't set on the Promotion object

src/components/promotion/PromotionDetails.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { CSSProperties, ReactElement } from 'react';
 import { HeartOutlined } from '@ant-design/icons';
-
 import { Promotion } from '../../types/promotion';
+import parse from '../../../node_modules/html-react-parser';
```
andrewzulaybar: Does this not work?
`import parse from 'html-react-parser';`

src/components/promotion/PromotionDetails.tsx
```diff
@@ -24,15 +23,17 @@ export default function PromotionDetails({
...
   return (
     <div style={styles.descriptionContainer}>
       <div style={styles.header}>
-        <h2>{name}</h2>
+        {boldName !== undefined ? parse(boldName) : <h2>{name}</h2>}
         <HeartOutlined style={styles.heart} />
       </div>
       <h3>{restaurantName}</h3>
-      <p>{description}</p>
+      {boldDescription !== undefined ? parse(boldDescription) : <p>{description}</p>}
```
andrewzulaybar: And then here you can probably just do `boldName ? parse(boldName) : <h2>{name}</h2>` and likewise for description

ericrkuo: Ohman idek what I was doing, good point thank you

ericrkuo:
@andrewzulaybar thank you so much! I'll take a look at your commit to understand the changes. All these reducers, providers and stuff are like magic to me, 😅

src/components/PromotionList.tsx
ericrkuo: Ah good catch, that's a good point.

src/components/navigation/SearchBar.tsx
```diff
@@ -13,16 +15,22 @@ const styles: { [identifier: string]: CSSProperties } = {
...
+      dropdownState.resetCallbacks.forEach((resetDropdown) => resetDropdown());
+
+      // Send search query
+      promotionsDispatch({
         type: DispatchAction.SEARCH_QUERY,
         payload: { searchQuery },
       });
     },
-    [dispatch]
+    [dropdownState.resetCallbacks, promotionsDispatch]
```
ericrkuo: What does it mean to add `dropdownState.resetCallbacks` to this parameter?

andrewzulaybar:
> @andrewzulaybar thank you so much! I'll take a look at your commit to understand the changes. All these reducers, providers and stuff are like magic to me, 😅

Ya for sure no problem! Let me know if there's something confusing or if you have a question, the reducers and providers and stuff definitely abstract a lot of things so I get tripped up all the time too

src/components/navigation/SearchBar.tsx
andrewzulaybar: Oh yeah this dependency array basically lets us tell the `useCallback` hook to update whenever any one of these variables get updated/changed. In this case, if `dropdownState.resetCallbacks` changes (e.g. another callback gets added) then the previously memorized version of this `onSearch` function would be replaced by a new one with the new `dropdownState.resetCallbacks` reference.

You usually don't have to think too much about what goes in here cause React tells you in the console if any dependencies it expects are missing!

ericrkuo: Ohhh that's so cool I see thanks for explaining

andrewzulaybar: No problem!

ericrkuo:
Would we want the search bar to clear when a user selects some filters? Right now we clear filters when we use search bar.

If we want this behaviour, would we need something like a SearchBarProvider and SearchBarReducer? Similar to the behaviour of where we have a dispatch action for dropdowns to reset the filters and a reset callback to set the active key

andrewzulaybar:
> Would we want the search bar to clear when a user selects some filters? Right now we clear filters when we use search bar.
> 
> If we want this behaviour, would we need something like a SearchBarProvider and SearchBarReducer? Similar to the behaviour of where we have a dispatch action for dropdowns to reset the filters and a reset callback to set the active key

Yeah this one I wasn't too sure about. From the user's perspective, I think it might be nice since the only way we're indicating that a user is seeing promotions related to the search term is if it's in the search bar. But from a dev perspective, it probably would require a little more time to get the infra setup. I think we should probably hold off on this for now unless it's something that you want to work on 😅

But yeah you're right! The SearchBarProvider would be needed to maintain + provide the search bar state for usage by other components.  And the reducer would probably have 2 actions SET_SEARCH_QUERY, REMOVE_SEARCH_QUERY or something like that to be able to manipulate the state

ericrkuo:
> I think we should probably hold off on this for now unless it's something that you want to work on 😅

Haha I think we can make it a separate ticket and I'll try to tackle it in another PR. But to continue the discussion, isn't there now a lot of coupling and less cohesion?

For example, the search bar component now needs to know to reset the filters and depends on the Dropdown context.
And vice versa, for when we add behaviour for clearing the search bar when a dropdown is selected.

andrewzulaybar:
> > I think we should probably hold off on this for now unless it's something that you want to work on 😅
> 
> Haha I think we can make it a separate ticket and I'll try to tackle it in another PR. But to continue the discussion, isn't there now a lot of coupling and less cohesion?
> 
> For example, the search bar component now needs to know to reset the filters and depends on the Dropdown context.
> And vice versa, for when we add behaviour for clearing the search bar when a dropdown is selected.

Ya if you wanna put it in the backlog that'd be great!

Yeah, I see what you mean -- that's a good point that I didn't really consider. Maybe for this specific task, yeah it might not be the best idea to introduce this dependency because it just makes the code more complicated. I think the traditional way of passing callbacks up/down the component hierarchy provides more of a "dependency-injection" type approach that makes it easy to abstract the dependency between the two components. 

More specifically, I'm thinking maybe the SearchBar could take a function prop like `onSearch` and it would be up to the parent component to define it. That way, the search bar isn't explicitly tied to the responsibility of resetting the dropdowns? And likewise for the dropdowns. I'm just not too sure off the top of my head where these components are relative to each other in the tree (and this would probably affect how easy it is to implement this)

ericrkuo:
Ohhh I see, that makes sense. I think for now since we don't have that many componentns, we can stick with this pattern. And I'm sure we can refactor a bunch later on once we have a better idea of how many components depend on each other and where they are relative in that ree

src/components/PromotionList.tsx
```diff
@@ -36,12 +36,17 @@ export default function PromotionList({
...
     filterPromotions(promotionListState.filter)
       .then((filteredPromotions: Promotion[]) =>
         sortPromotions(filteredPromotions, promotionListState.sort)
       )
       .then((sortedPromotions: Promotion[]) => {
         dispatch({ type: DispatchAction.DATA_SUCCESS });
+
+        // If search bar contains keyword, reset it. This is only minimal behaviour, needs work with PP-68
+        dispatch({ type: DispatchAction.RESET_SEARCH_QUERY });
+
```
ericrkuo: I added this temporary workaround to handle the case where you do 
1. search bar "fry"
2. choose filter/dropdown
3. search bar "fry" again (same searchQuery)
This ensures that for 3, it gets properly called because before, since the searchQuery didn't change, it actually wouldn't have gotten the promotions.

src/components/PromotionList.tsx
```diff
@@ -36,12 +36,17 @@ export default function PromotionList({
    */
   useEffect(() => {
     dispatch({ type: DispatchAction.DATA_LOADING });
```
ericrkuo, line 38: Does this dispatch action need to be outside of the useEffect? Or do we need another one of these dispatch actions in the useEffect for the search query

src/components/PromotionList.tsx
andrewzulaybar: Ohh I see, that makes sense good catch!!

src/components/PromotionList.tsx
andrewzulaybar: I think I put this inside the useEffect to know when the promotions are being fetched (and in the future this would be used to determine when a Loading indicator would be displayed). It isn't being used right now but I thought I'd set it up anyways, so yeah maybe if you can add one to the if-block for the search query that would be nice!

src/components/PromotionList.tsx
```diff
@@ -50,6 +55,18 @@ export default function PromotionList({
       });
   }, [promotionListState.filter, promotionListState.sort, dispatch]);
 
+  /**
+   * When a search query is set, fetches promotions that satisfy the query.
+   */
+  useEffect(() => {
+    if (promotionListState.searchQuery) {
+      getPromotions([{ searchQuery: promotionListState.searchQuery }]).then((promotions) => {
```
andrewzulaybar, line 64: If you could add the dispatch call for DATA_LOADING before this line that would be great!
```suggestion
      dispatch({ type: DispatchAction.DATA_LOADING });
      getPromotions([{ searchQuery: promotionListState.searchQuery }]).then((promotions) => {
```

andrewzulaybar (approved):
Just one comment but otherwise everything looks great :)

PR #16, opened 2021-02-09
PP-10 Addressed NOT_FOUND Error in Google Places API

lilyydu:
We may encounter the `NOT_FOUND` error when we issue a Places Detail request for a specific restaurant. This will require us to make a Place Details - ID Refresh SKU request with only the ID field in the parameter. 

We may encounter `NOT_FOUND` from the refresh request as well- which means we will need to get the restaurant name and location from the BE and make a Places Search again through `getRestaurantInfo`. This part is not addressed in this PR- it will be wherever we're making the call to fetch info for a particular restaurant.

I'm currently making changes in the BE, once that's up, I'll link it here as well.


test/services/GooglePlacesService.test.ts
```diff
@@ -52,7 +52,7 @@ describe('Unit tests for GooglePlacesService', function () {
         expect(result).toHaveProperty('name', 'RIB & CHICKEN');
         expect(result).toHaveProperty('opening_hours');
         expect(result).toHaveProperty('photos');
-        expect(result).toHaveProperty('rating', 4.5);
```
lilyydu, line 57: I removed these numerical values in case we add CI later for our tests- these numbers change pretty often

src/services/GooglePlacesService.ts
```diff
@@ -87,7 +87,14 @@ class GooglePlacesService {
...
       })
       .then((response: AxiosResponse) => {
-        const restaurantResults = response?.data?.result;
+        const restaurantData = response?.data;
+        const restaurantResults = restaurantData?.result;
+
+        // check if placeID is valid
+        if (restaurantData.status && restaurantData.status === 'NOT_FOUND') {
+          return this.refreshPlaceIDAndGetDetails(placeID);
+        }
```
ericrkuo: What scenario would `response` be undefined?

Can we just check for `response.status === 400`?

src/services/GooglePlacesService.ts
```diff
@@ -87,7 +87,14 @@ class GooglePlacesService {
         },
       })
       .then((response: AxiosResponse) => {
-        const restaurantResults = response?.data?.result;
+        const restaurantData = response?.data;
+        const restaurantResults = restaurantData?.result;
+
+        // check if placeID is valid
+        if (restaurantData.status && restaurantData.status === 'NOT_FOUND') {
```
ericrkuo: Also if status is `NOT_FOUND`, will that go to the catch block or then block

src/services/GooglePlacesService.ts
```diff
@@ -113,6 +120,31 @@ class GooglePlacesService {
...
+        params: {
+          place_id: placeID,
+          fields: 'place_id',
+          key: process.env.REACT_APP_GOOGLE_PLACES_API_KEY,
+        },
+      })
+      .then((response: AxiosResponse) => {
+        const data = response?.data;
+
+        if (data.status && data.status === 'NOT_FOUND') {
```
ericrkuo: Same comment here

src/services/GooglePlacesService.ts
```diff
@@ -113,6 +120,31 @@ class GooglePlacesService {
...
+          return Promise.reject(new Error('NOT_FOUND'));
+        }
```
ericrkuo: How come we want to reject with error?

src/services/GooglePlacesService.ts
```diff
@@ -131,6 +163,17 @@ class GooglePlacesService {
         return Promise.reject(err);
       });
   }
+
+  // helper function used to refresh place ID and call getRestaurantDetails again for NOT_FOUND error
+  refreshPlaceIDAndGetDetails(placeID: string): Promise<RestaurantDetails> {
+    return this.refreshPlaceID(placeID)
+      .then((newPlaceID: string) => {
+        return this.getRestaurantDetails(newPlaceID);
+      })
```
ericrkuo: Hmm, I'm kind of scared of this recursive call resulting in a stack overflow error

test/services/GooglePlacesService.test.ts
```diff
@@ -68,6 +68,17 @@ describe('Unit tests for GooglePlacesService', function () {
       });
   });
 
+  test('Get details for a place using invalid placeID, should be unsuccessful', () => {
+    return googlePlacesAPI
+      .getRestaurantDetails('ChIJ2fF6zviX9YgRflKEZNrpLrQ')
```
ericrkuo: Could we turn `ChIJ2fF6zviX9YgRflKEZNrpLrQ` and `ChIJb0n5cWl3hlQRIbVGYLiTEgE` into constants with variable names that represent if they're valid or invalid placeIds?

test/services/GooglePlacesService.test.ts
```diff
@@ -137,4 +148,56 @@ describe('Unit tests for GooglePlacesService', function () {
         fail('Did not expect to fail: ' + error.message);
       });
   });
+
+  test('Refreshing placeID, should be successful', () => {
```
ericrkuo:
```suggestion
  test('Refreshing a valid placeID should be successful', () => {
```

test/services/GooglePlacesService.test.ts
```diff
@@ -137,4 +148,56 @@ describe('Unit tests for GooglePlacesService', function () {
...
+      .refreshPlaceID('ChIJb0n5cWl3hlQRIbVGYLiTEgE')
+      .then((result: string) => {
+        expect(result).toEqual('ChIJb0n5cWl3hlQRIbVGYLiTEgE');
+      })
+      .catch((error: AxiosError) => {
+        fail('Did not expect to fail: ' + error.message);
+      });
+  });
+
+  test('Refreshing placeID, should be unsuccessful', () => {
```
ericrkuo:
```suggestion
  test('Refreshing an invalid placeID should be unsuccessful', () => {
```

src/services/GooglePlacesService.ts
```diff
@@ -87,7 +87,14 @@ class GooglePlacesService {
         },
       })
       .then((response: AxiosResponse) => {
-        const restaurantResults = response?.data?.result;
+        const restaurantData = response?.data;
+        const restaurantResults = restaurantData?.result;
```
ericrkuo: Sort of an unrelated note, 
https://github.com/ubclaunchpad/PromoPal-frontend/blob/f585ec71668c22d4753096e7dcba970e20e665f8/src/services/GooglePlacesService.ts#L73-L75
How come we need to cast it? the value type is already `RestaurantDetails` right?

ericrkuo: Also could we make `currRestaurants` private?

src/services/GooglePlacesService.ts
```diff
@@ -113,6 +120,31 @@ class GooglePlacesService {
...
 
+  // refreshes the invalid placeID, may also result in NOT_FOUND error
+  refreshPlaceID(placeID: string): Promise<string> {
+    return axios
+      .get('https://maps.googleapis.com/maps/api/place/details/json?', {
+        params: {
+          place_id: placeID,
+          fields: 'place_id',
+          key: process.env.REACT_APP_GOOGLE_PLACES_API_KEY,
+        },
```
ericrkuo, line 132: How come the `refresh URL` is the same as URL in `getRestaurantDetails`?

If `getRestaurantDetails` resulted in a NOT_FOUND won't this one be guaranteed to be a NOT_FOUND as well, since the URL is the exact same?

ericrkuo: Just kidding, maybe we can write an explanation for other developers summarizing this point below?
https://developers.google.com/places/web-service/place-id#save-id

src/services/GooglePlacesService.ts
ericrkuo: Ah so this is like when the googles refresh fails refreshing :P

src/services/GooglePlacesService.ts
richardyjtian: ^ I was looking at this and think it should be private if possible too.

richardyjtian (commented):
This looks really good! I had a question about dotenv though - do we need to import it because GooglePlacesService is a class? I remember being able to just reference process.env.REACT_APP_GOOGLE_MAPS_API_KEY in other components.

src/services/GooglePlacesService.ts
andrewzulaybar: I think one way we could approach handling this case is by changing the signature of `refreshPlaceID` to include a `handleNotFoundError` optional parameter which would be a callback. And then we can extract

```
 if (restaurantData.status && restaurantData.status === 'NOT_FOUND') {
    return this.refreshPlaceIDAndGetDetails(placeID);
 }
```

into a private helper function and replace it with a call to that helper function (and also we would use this function as a default for this parameter). Then here on line 171, we could just pass in a callback that doesn't include a call to `refreshPlaceIDAndGetDetails`, terminating the recursion?

src/services/GooglePlacesService.ts
```diff
@@ -131,6 +163,17 @@ class GooglePlacesService {
...
+  refreshPlaceIDAndGetDetails(placeID: string): Promise<RestaurantDetails> {
+    return this.refreshPlaceID(placeID)
+      .then((newPlaceID: string) => {
+        return this.getRestaurantDetails(newPlaceID);
+      })
+      .catch((err: AxiosError) => {
+        return Promise.reject(err);
+      });
+  }
 }
```
andrewzulaybar, line 181: For this class, could you include visibility keywords at the beginning of each function? I think right now they're all implicitly public and I'm not sure if you wanted them to be!

test/services/GooglePlacesService.test.ts
andrewzulaybar: Good catch!

andrewzulaybar (changes requested):
Nice job! Just a few comments to resolve -- let us know when you get to them :)

src/services/GooglePlacesService.ts
lilyydu: Are you referring to `response: AxiosResponse` ?

src/services/GooglePlacesService.ts
lilyydu: Nope, thought that was weird too but surprisingly goes to the `then` block

src/services/GooglePlacesService.ts
lilyydu: We need to cast it cause there's a type error that it could be undefined

EDIT:
I'll make the change to private!

src/services/GooglePlacesService.ts
lilyydu: Which point did you want me to add in particular?

lilyydu:
> This looks really good! I had a question about dotenv though - do we need to import it because GooglePlacesService is a class? I remember being able to just reference process.env.REACT_APP_GOOGLE_MAPS_API_KEY in other components.

Ahh yeah I made this change awhile ago but it might've messed up when I merged. Nice catch tho, I'll make the change!

lilyydu:
Addressed comments and ready for re-review!

src/services/GooglePlacesService.ts
ericrkuo: Ohh I see, what scenario would it go to the catch block?

src/services/GooglePlacesService.ts
ericrkuo: Oh like for me it wasn't trivial to see that even though the URL is the same, when you only specify the `place_id` it does the refresh. But if you add more fields, then it won't do the refresh/

image
^Keyword **only**

src/services/GooglePlacesService.ts
ericrkuo: Oh but I think if you cast it and it is undefined, it will still be undefined
image

src/services/GooglePlacesService.ts
ericrkuo: ya

lilyydu: Are you referring to the `?` in `const restaurantData = response?.data;`

We can't check for `response.status === 400` because it explicitly returns a `NOT_FOUND` status code when the place ID is invalid which is the error we want to handle for this PR. To add, it actually returns a status 200- even though this seems like it should be 400.

I couldn't find a specific list of status codes for Google Places API- this one is for Android vers but here are some [examples](https://developers.google.com/places/android-sdk/reference/com/google/android/libraries/places/api/net/PlacesStatusCodes)

ericrkuo: Ohh I see, I thought the `NOT_FOUND` was 400 status code. 

But my other question is, why do we need `response?.data`, could it be `response.data`

src/services/GooglePlacesService.ts
lilyydu: I wasn't able to encounter or find specific documentation on it but one would definitely be [usage limits](https://developers.google.com/maps/premium/previous-licenses/articles/usage-limits)

src/services/GooglePlacesService.ts
lilyydu: Are you suggesting to do another conditional check inside, to check that its not undefined?

src/services/GooglePlacesService.ts
lilyydu: Hmm.. I guess it could be- but I wanted to be safe! I couldn't find any particular scenarios where past users were returned `undefined` for Places API but it's happened for Maps API and Nearby Search, etc so I think it's better safe than sorry unless it's harmful to include the extra check?

src/services/GooglePlacesService.ts
richardyjtian: Wait, if the placeID is valid, this.currRestaurants.get(placeID) will always return a defined RestaurantDetails, so I don't think it's necessary to have another check. But yeah I see why you would need to cast it, because "get" doesn't know if the placeID's valid.

richardyjtian (approved):

src/services/GooglePlacesService.ts
ericrkuo: Ah shoot forgot there is an if statement that checks the existence of the key. Sorry about that

src/services/GooglePlacesService.ts
ericrkuo: Maps API and Nearby Search returned undefined responses from axios? Isn't that very unlikely because the signature for `axios.get` always returns the interface `AxiosResponse`?

src/services/GooglePlacesService.ts
```diff
@@ -81,7 +85,16 @@ class GooglePlacesService {
         },
       })
       .then((response: AxiosResponse) => {
-        const restaurantResults = response?.data?.result;
+        const restaurantData = response?.data;
+        const restaurantResults = restaurantData?.result;
+
+        // check if placeID is valid
+        if (restaurantData.status && restaurantData.status === 'NOT_FOUND') {
```
andrewzulaybar: I just found out that you can use optional chaining to eliminate the first condition!
 
```suggestion
        if (restaurantData?.status === 'NOT_FOUND') {
```

src/services/GooglePlacesService.ts
```diff
@@ -107,8 +120,33 @@ class GooglePlacesService {
...
+        params: {
+          place_id: placeID,
+          fields: 'place_id',
+          key: process.env.REACT_APP_GOOGLE_PLACES_API_KEY,
+        },
+      })
+      .then((response: AxiosResponse) => {
+        const data = response?.data;
+
+        if (data.status && data.status === 'NOT_FOUND') {
```
andrewzulaybar: Same as above

```suggestion
        if (data?.status === 'NOT_FOUND') {
```

andrewzulaybar (approved):
This looks great! Just added some minor suggestions :)

PR #1, opened 2021-01-30
Adding CI for FE

ericrkuo:
https://github.com/ubclaunchpad/PromoPal-backend/issues/92

PR #20, opened 2021-02-21
feat(promotions): Implement sorting of promotions

andrewzulaybar:
- Created the Geometry Service for computation of distances
- Implemented sorting functionality for the dropdown options: Distance, Popularity, and Recency

**Questions:**
- Is # of votes how we plan to determine the popularity of a promotion?

**Related ticket(s):**
- https://promopal.atlassian.net/browse/PP-11


ericrkuo:
> Is # of votes how we plan to determine the popularity of a promotion?

Yes, that was what I was thinking for the time being. We could actually do a more complex algorithm if we wanted to that would factor in the following (not in a specific order)
1. Number of votes
2. The date the promotion was created
3. Data that we could start tracking in order to better sort by popularity
    a. Number of views per certain duration of time
    b. Number of votes per certain duration of time
    c. If we do allow the making comments feature, we could extract sentimental details from a promotion (positive promotions would have people agreeing that this promo/restaurant is good, therefore more popular)

src/services/PromotionService.ts
```diff
@@ -108,33 +111,56 @@ export function filterPromotions(filters: FilterOptions): Promise<Promotion[]> {
...
+      reject(promotions);
+    }
+  });
 }
 
-// TODO: see https://github.com/ubclaunchpad/foodies/issues/99
 function sortByPopularity(promotions: Promotion[]) {
-  return promotions;
+  return promotions.sort((a, b) => b.votes - a.votes);
 }
```
ericrkuo, line 155: Wouldn't we need to support the upvote/downvote functionality in order to sort by popularity?

ericrkuo:
Now that I'm thinking about it, should sorting be in the BE?

Cause imagine we have 1000 promotions, there's no way we will show all 1000 on the home page immediately, that would take too much time right. Like we could show 50, but then if we want to sort, that means we would only be sorting 50 and disregarding the other 950 promotions. We could delegate sorting functionality to BE. What do you think?

andrewzulaybar:
> > Is # of votes how we plan to determine the popularity of a promotion?
> 
> Yes, that was what I was thinking for the time being. We could actually do a more complex algorithm if we wanted to that would factor in the following (not in a specific order)
> 
> 1. Number of votes
> 2. The date the promotion was created
> 3. Data that we could start tracking in order to better sort by popularity
>    a. Number of views per certain duration of time
>    b. Number of votes per certain duration of time
>    c. If we do allow the making comments feature, we could extract sentimental details from a promotion (positive promotions would have people agreeing that this promo/restaurant is good, therefore more popular)

Ahh those are good points! I think `Number of votes per certain duration of time` would be a good tradeoff between having an effective but simple solution? I think you're right, just having number of votes would degrade over time to just being "Most Votes" instead of "Most Popular".

So maybe we could do something like most number of votes in the past two weeks? past month?

I also wasn't sure if we wanted to factor in # of saves by different users?

src/services/PromotionService.ts
andrewzulaybar: Yeah, you're right so currently this wouldn't do anything. I can add a ticket to work on upvoting/downvoting in the next sprint?

andrewzulaybar: https://promopal.atlassian.net/browse/PP-71

ericrkuo: Yup sure sounds good!

andrewzulaybar:
> Now that I'm thinking about it, should sorting be in the BE?
> 
> Cause imagine we have 1000 promotions, there's no way we will show all 1000 on the home page immediately, that would take too much time right. Like we could show 50, but then if we want to sort, that means we would only be sorting 50 and disregarding the other 950 promotions. We could delegate sorting functionality to BE. What do you think?

Yeah... you're right sorry didn't think about that too much :( I guess currently because the BE is returning the promotions arbitrarily then it wouldn't work. Also kind of related, but do you think we should implement some sort of pagination for promotions on the BE?

ericrkuo:
> Ahh those are good points! I think `Number of votes per certain duration of time` would be a good tradeoff between having an effective but simple solution? I think you're right, just having number of votes would degrade over time to just being "Most Votes" instead of "Most Popular".
> 
> So maybe we could do something like most number of votes in the past two weeks? past month?
> 
> I also wasn't sure if we wanted to factor in # of saves by different users?

Ah yeah most number of votes in the past X weeks/months sounds like a good idea, but BE would need some sort of refactoring and logging capability to know this. Factoring in # of saves by different users is also a very very good idea, I like this one a lot. Created a ticket here https://promopal.atlassian.net/browse/PP-72 to start working on this, most likely this would need to be in BE as well

ericrkuo:
> > Now that I'm thinking about it, should sorting be in the BE?
> > Cause imagine we have 1000 promotions, there's no way we will show all 1000 on the home page immediately, that would take too much time right. Like we could show 50, but then if we want to sort, that means we would only be sorting 50 and disregarding the other 950 promotions. We could delegate sorting functionality to BE. What do you think?
> 
> Yeah... you're right sorry didn't think about that too much :( I guess currently because the BE is returning the promotions arbitrarily then it wouldn't work. Also kind of related, but do you think we should implement some sort of pagination for promotions on the BE?

No need to apologize, I think I remember we talked about it long time ago and thought FE would be best to do the sorting :P. What do you mean by pagination for promotions on the BE?

andrewzulaybar:
> Ah yeah most number of votes in the past X weeks/months sounds like a good idea, but BE would need some sort of refactoring and logging capability to know this. Factoring in # of saves by different users is also a very very good idea, I like this one a lot. Created a ticket here https://promopal.atlassian.net/browse/PP-72 to start working on this, most likely this would need to be in BE as well

Yay thanks! Yeah, I think I need to think about this one a bit more too cause I guess it could become a bit more complex...

andrewzulaybar:
> No need to apologize, I think I remember we talked about it long time ago and thought FE would be best to do the sorting :P. What do you mean by pagination for promotions on the BE?

Yeah I remember us deciding on that while we thought filtering would be delegated to the BE, not sure why I didn't think to put sorting on the BE either 😅  and by pagination for promotions, like instead of returning all 1000 promotions to the FE, the FE could request a certain # of promotions at a time (e.g. 50). I made a ticket for it on the FE this past sprint but I don't think it considers the case you mentioned: https://promopal.atlassian.net/browse/PP-22

I think this kind of factors in sorting/filtering because like if we wanted to sort promotions, we should probably sort first and then fetch promotions #1 - #50. And to get the second page we'd sort again and fetch #51 - #100 etc? What do you think

ericrkuo:
> Yeah I remember us deciding on that while we thought filtering would be delegated to the BE, not sure why I didn't think to put sorting on the BE either 😅 and by pagination for promotions, like instead of returning all 1000 promotions to the FE, the FE could request a certain # of promotions at a time (e.g. 50). I made a ticket for it on the FE this past sprint but I don't think it considers the case you mentioned: https://promopal.atlassian.net/browse/PP-22
> 
> I think this kind of factors in sorting/filtering because like if we wanted to sort promotions, we should probably sort first and then fetch promotions #1 - #50. And to get the second page we'd sort again and fetch #51 - #100 etc? What do you think

Ohya I think we should definitely do pagination, but I have no clue how to implement that. I guess we can research and do that in https://promopal.atlassian.net/browse/PP-22, but our first priority would be doing the sorting in the BE and BE would always return a limit of 50 promotions? Then after we're finished that in both the BE and FE, we can work on PP-22?

ericrkuo:
I created a ticket https://promopal.atlassian.net/browse/PP-73 to do sorting in the BE. Does this mean we should close this PR, and come back to it until the BE has supported functionality for sorting?

andrewzulaybar:
> Ohya I think we should definitely do pagination, but I have no clue how to implement that. I guess we can research and do that in https://promopal.atlassian.net/browse/PP-22, but our first priority would be doing the sorting in the BE and BE would always return a limit of 50 promotions? Then after we're finished that in both the BE and FE, we can work on PP-22?

Yeah I'm not too sure about best practices yet and I'll need to do a little research too, but I was thinking off the top of my head something like if the user goes to the URL ending with `/promotions/2` then we'd make a request to a BE endpoint like `promotions/:pageNum` and it would know to get the promotions ranging from pageNum*NUM_PROMOTIONS_PER_PAGE => (pageNum + 1)*NUM_PROMOTIONS_PER_PAGE

Yeah, sorting would go first and then this could be anytime after!

andrewzulaybar:
> I created a ticket https://promopal.atlassian.net/browse/PP-73 to do sorting in the BE. Does this mean we should close this PR, and come back to it until the BE has supported functionality for sorting?

Thanks! And yeah, I'll close this PR and rebase my other ones!

ericrkuo:
Oh, @andrewzulaybar did you want to do sorting in the BE? I wouldn't mind doing it either

andrewzulaybar:
> Oh, @andrewzulaybar did you want to do sorting in the BE? I wouldn't mind doing it either

Actually I think I can do it since I haven't done much work with the BE! Would be a good way for me to get to know the codebase more :)

PR #21, opened 2021-02-21
feat(map): Center map on user's current location

andrewzulaybar:
- Created LocationService to handle fetching of current location using w3c's Geolocation API
- Centers the map on the user's current location after the initial page load
- Displays markers on map for the promotions currently listed

**Related ticket(s):**
- https://promopal.atlassian.net/browse/PP-17
- https://promopal.atlassian.net/browse/PP-18
- https://promopal.atlassian.net/browse/PP-19

### DIFF
--- a/src/services/LocationService.ts
+++ b/src/services/LocationService.ts
@@ -1,13 +1,13 @@
 // Types taken from the w3c Geolocation API: https://w3c.github.io/geolocation-api/#geolocation_interface
 
 export interface GeolocationCoordinates {
-  accuracy: number;
-  altitude: number | null;
-  altitudeAccuracy: number | null;
-  heading: number | null;
+  accuracy?: number;
+  altitude?: number | null;
+  altitudeAccuracy?: number | null;
+  heading?: number | null;
   latitude: number;
   longitude: number;
-  speed: number | null;
+  speed?: number | null;
 }
 
 export interface GeolocationPosition {
@@ -23,6 +23,16 @@ export interface GeolocationPositionError {
 
 class LocationService {
   /**
+   * Default location is downtown Vancouver
+   */
+  public get defaultLocation(): GeolocationPosition {
+    return {
+      coords: { latitude: 49.282, longitude: -123.1171 },
+      timestamp: Date.now(),
+    };
+  }
+
+  /**
    * Retrieves the user's current location using navigator.geolocation.
    * Times out and calls the error callback after 3s.
    */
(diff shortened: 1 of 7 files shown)


src/types/promotion.ts
```diff
@@ -19,7 +21,9 @@ export interface Promotion {
   lon: number;
   liked: boolean;
   name: string;
+  placeId: string;
   promotionType: string;
+  restaurant: RestaurantDetails;
   restaurantName: string;
```
ericrkuo, line 26: What's `RestaurantDetails`?

src/services/LocationService.ts
```diff
@@ -0,0 +1,44 @@
...
+
+class LocationService {
+  /**
+   * Retrieves the user's current location using navigator.geolocation.
+   * Times out and calls the error callback after 3s.
+   */
+  public getCurrentLocation(): Promise<GeolocationPosition> {
+    return new Promise((resolve, reject) => {
+      if (navigator.geolocation) {
+        const onSuccess = (position: GeolocationPosition) => resolve(position);
```
ericrkuo, line 42: Ah so this will prompt user for location access too right

src/services/LocationService.ts
```diff
@@ -0,0 +1,44 @@
...
+        const onError = (error: GeolocationPositionError) => reject(error);
+        const options = { timeout: 3000 };
+        navigator.geolocation.getCurrentPosition(onSuccess, onError, options);
+      } else {
+        const error = new Error('navigator.geolocation is not defined');
+        reject(error);
+      }
+    });
+  }
+}
```
ericrkuo: Also if we reach the `else` block or the `onError` function, rather than rejecting with an err, could we instead return the default map location `const defaultCenter = { lat: 49.282, lng: -123.1171 };` in MapContainer.tsx?

src/types/promotion.ts
andrewzulaybar: Oh it's the interface that Lily created for the return value from Google Places Details -- in #22 I think I refactor this to just be `Restaurant` which is a superset of the properties of `RestaurantDetails`

src/services/LocationService.ts
andrewzulaybar: Yeah! It should show up as the info window in your browser that says something like "Do you allow this website to access your location?"

src/services/LocationService.ts
andrewzulaybar: Yeah, good point I can do that!

src/services/LocationService.ts
ericrkuo: Ahh got it kk thanks

ericrkuo (approved):

PR #22, opened 2021-02-21
PP-20: fix(map): Allow markers to open restaurant card

andrewzulaybar:
- Added an event listener to the Google Map Markers such that it opens up the restaurant card on click
- Refactored the `PromotionCard` component to take an `onClick` prop because the onClick action will vary depending on where we use the component (e.g., we don't want to toggle the RestaurantCard if we're on the "My Promotions" page)
- Extracted the action of toggling the restaurant card into the RestaurantCardReducer, which made it easier to define the `onClick` actions for the PromotionCard and the Google Map Markers
- Moved the `RestaurantDetails` interface into the `types/restaurants.ts` file @ld357

**Related ticket(s):**
- https://promopal.atlassian.net/browse/PP-20


src/services/PromotionService.ts
```diff
@@ -50,29 +50,25 @@ export async function getPromotions(query?: PromotionDTO[]): Promise<Promotion[]
     });
 }
 
-export async function getRestaurant({ id }: Promotion): Promise<Restaurant> {
+export async function getRestaurant(id: string): Promise<Restaurant> {
   // TODO: find restaurant given placeId from promotion
   return {
```
ericrkuo: Do we have a ticket to track this TODO?

ericrkuo: And also this would be the area where we could store the restaurant details in the session data? So that for the same user, if they click on the marker again, we don't need to call google places API again right

ericrkuo (approved):

src/services/PromotionService.ts
andrewzulaybar: Oh yeah, I'll link it in the code: https://promopal.atlassian.net/browse/PP-25

And I think that caching might already be done in the `GooglePlacesService`! This function would eventually call `GooglePlacesService.getRestaurantDetails` which checks the cache if it already has the details for this restaurant

ericrkuo: Oh wait you're right sorry forgot that we cache it.

lilyydu: I opened this again since we've changed it now 😅  so you might need to add more details for user session?

lilyydu (approved):
I couldn't fully verify the FE but the functionality looked right!

src/services/PromotionService.ts
andrewzulaybar: Ahhh thanks! Just opened up a ticket in the backlog for this: https://promopal.atlassian.net/browse/PP-84

I'm thinking this could be an improvement? When we fetch the restaurant, I think it should be attached to a promotion on the FE so that we won't need to make the additional request again (we'd just do `promotion.restaurant` if we need it again) but I think the ticket will be nice as a reminder to check that we're not making the additional requests!

lilyydu: Oooh gotcha! Would that be enough for user state- so same idea as the hashmap we had previously?

andrewzulaybar: Actually yeah you're right, might be better to do it as the hashmap like you had previously so then it's accessible throughout the whole app in case we need it 😛

PR #23, opened 2021-02-21
feat(my-promotions): Fetch user's uploaded promotions

andrewzulaybar:
- Created UserService to handle requests related to the `/users` endpoint
- Added an `ApiResonse` type to type resource requests from the backend
- Created an `isError` util function so that we have a more uniform way to check whether the responses are errors or not

**Related ticket(s):**
- https://promopal.atlassian.net/browse/PP-51

### DIFF
--- a/src/screens/MyPromotions.tsx
+++ b/src/screens/MyPromotions.tsx
@@ -1,11 +1,12 @@
 import { Checkbox, Col, Row } from 'antd';
-import React, { CSSProperties, ReactElement } from 'react';
+import React, { CSSProperties, ReactElement, useEffect, useState } from 'react';
 
 import UploadPromoButton from '../components/button/UploadPromoButton';
 import DropdownMenu from '../components/DropdownMenu';
 import PromotionCard from '../components/promotion/PromotionCard';
+import UserService from '../services/UserService';
 import { Dropdown, DropdownType } from '../types/dropdown';
-import { Promotion, Schedule, User } from '../types/promotion';
+import { Promotion, User } from '../types/promotion';
 
 const dropdowns: Dropdown[] = [
   {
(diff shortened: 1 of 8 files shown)


src/screens/MyPromotions.tsx
```diff
@@ -62,94 +63,14 @@ const dropdowns: Dropdown[] = [
 ];
 
 const user: User = {
-  id: 'u1',
+  id: '8f8fc016-5bb2-4906-ad88-68932c438665',
```
ericrkuo: Ah so this will only work if everyone has a user with `id: '8f8fc016-5bb2-4906-ad88-68932c438665'`. I guess this is temporary until we have user auth implemented

src/services/UserService.ts
```diff
@@ -0,0 +1,22 @@
+import { UploadedPromotionsResponse } from '../types/api';
+import { Promotion } from '../types/promotion';
+import { isError } from '../utils/api';
+import Routes from '../utils/routes';
+
+class UserService {
+  public async getUploadedPromotions(userId: string): Promise<Promotion[]> {
+    const url = `${Routes.USERS}/${userId}/uploadedPromotions/`;
```
ericrkuo: Minor thing, but how about

```suggestion
    const url = `${Routes.USERS}/${userId}/${Routes.UPLOADED_PROMOTIONS}/`;
```

src/services/UserService.ts
```diff
@@ -0,0 +1,22 @@
...
+    return fetch(url)
+      .then((response: Response) => response.json())
+      .then((response: UploadedPromotionsResponse) => {
+        if (isError<UploadedPromotionsResponse>(response)) {
+          return Promise.reject(response);
+        }
+        const { uploadedPromotions } = response;
+        return Promise.resolve(uploadedPromotions);
```
ericrkuo, line 16: Also, I think we should talk about how much detail we want to show in the uploaded and saved promotions page compared to the main section. 

I wasn't really too sure if for uploaded and saved promotions we want to return the discounts and schedules as well

src/types/api.ts
```diff
@@ -0,0 +1,16 @@
+import { Promotion } from './promotion';
+import { User } from './user';
+
+// Generic
+export type ApiResponse<T> = T | ApiError;
+
+export interface ApiError {
+  errorCode: string;
+  message: string[];
+}
```
ericrkuo, line 10: Awesome this is really nice

src/screens/MyPromotions.tsx
andrewzulaybar: Yeah, I couldn't think of another solution for now. I just took this id from the init_data.sql file, so hopefully everyone should have it if they have their db up-to-date? Not sure how often this file changes as well

src/services/UserService.ts
andrewzulaybar: Oooh you gave me an idea -- what about:

in routes.ts:
```
USERS: {
  UPLOADED_PROMOTIONS: (id: string): string => `/users/${id}/uploadedPromotions`,
},
```

and in this file it would be:
```
Routes.USERS.UPLOADED_PROMOTIONS(userId);
```

Not sure though, what do you think?

src/services/UserService.ts
andrewzulaybar: Ah good point, I didn't notice that. From the design, I just used the regular PromotionCard component that we display on the home page so I think it just uses schedules?

Screen Shot 2021-02-21 at 1 50 23 PM

src/services/UserService.ts
ericrkuo: I like this idea! I didn't know you could specify arrow functions like that, really neat!

src/services/UserService.ts
ericrkuo: Oh I see, hmmm right now I noticed that for BE we return the user information and we also don't include the schedules or discounts. I think we can just return an array of promotions joined with the discounts/schedules.

ericrkuo (approved):

PR #24, opened 2021-02-21
feat(image-upload): Implement image selector and image upload to S3

richardyjtian:
Image uploading functionality-wise is complete
-
nofile
-
imageselected
-
onupload
-
uploadedimage

Future work on this feature includes styling ImageSelector.tsx, integrating the upload image button with the UploadPromotion page, restricting API calls to deployed promopal hostname, only allowing authenticated users to upload, and setting a limit on uploads from the same IP address.


src/components/ImageSelector.tsx
```diff
@@ -0,0 +1,40 @@
...
+      {image.imageBinary && <img src={image.imageBinary} alt="" />}
+      {/* Displays the remove image button if an image is selected */}
+      {image.imageBinary && (
+        <button onClick={() => setImage({ imageBinary: '', imageType: '' })}>Remove image</button>
+      )}
+      {/* Displays the upload image button if an image is selected */}
+      {/* TODO: This will need to be removed in the future, as the upload promotion button
+          in the upload promotion page will invoke AmazonS3Service.uploadImage(image, promotionId) 
+          after an promotion has been successfully saved to our backend db
+      */}
```
ericrkuo: Good point, could we track this TODO in a ticket and link the ticket to this comment?

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,75 @@
...
+    });
+  }
+
+  uploadImage(image: Image, promotionId: string) {
+    // Get the presigned URL
+    axios
+      .get(process.env.REACT_APP_AWS_PRESIGNED_URL_ENDPOINT as string, {
+        params: {
+          promotionId: promotionId,
+          imageType: image.imageType,
```
ericrkuo: What's the `promotionId` key, is that the name/key of the object?
I thought it would be something like:
```
          Key: promotionId,
          imageType: image.imageType,
```

ericrkuo (commented):
Great work, just some quick questiosn. For your comment about the future work, could we create these all as tickets under the JIRA board?

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,75 @@
+import axios, { AxiosError, AxiosResponse } from 'axios';
+
+interface Image {
```
lilyydu: Not sure if we'll plan on accessing this interface in other files but if so, maybe extract this into its own class?

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,75 @@
...
+    return new Promise((resolve, reject) => {
+      reader.onload = (e: ProgressEvent<FileReader>) => {
+        // Only allow jpg, jpeg, png, and gifs to be selected
+        const file = e?.target?.result as string;
+        const fileRegexMatches = file.match(/image\/(jpe?g|png|gif)/);
+        if (fileRegexMatches === null) {
+          return reject(new Error('Wrong file type. Only images are allowed.'));
+        }
+        // Restrict the size of the file
+        if (file.length > parseInt(process.env.REACT_APP_AWS_MAX_IMAGE_SIZE as string)) {
```
lilyydu: Should we also have a minimum image size?

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,75 @@
+import axios, { AxiosError, AxiosResponse } from 'axios';
+
+interface Image {
+  imageBinary: string;
+  imageType: string;
```
lilyydu: Could we restrict the types here somehow? Through an enum or something similiar!

lilyydu (commented):
Great job! Just small comments, also wondering if we're planning to add captions to our photos, that might be another ticket to add? There's also a warning on line 61 in `AmazonS3Service`- not sure if you wanted to return result or just keep it as that

src/components/ImageSelector.tsx
richardyjtian: yes! I've created PP-75 and PP-76 to track these [https://promopal.atlassian.net/browse/PP-595](url)

src/services/AmazonS3Service.ts
richardyjtian: promotionId is just the promotionId of the image's promotion; in s3, the key = promotionId.imageType (e.g. 12345.jpg)

src/services/AmazonS3Service.ts
richardyjtian: yes, that's a good point, but I don't think it's necessary to use this interface outside of AmazonS3Service and ImageSelector; everywhere else we can just use href=s3imagelink

src/services/AmazonS3Service.ts
richardyjtian: I think it should be okay; I think problems only arise when a user tries to upload too large an image

src/services/AmazonS3Service.ts
richardyjtian: sorry I'm not quite sure what you mean :sweat_smile:, but I think I've restricted the types to strings right?

src/services/AmazonS3Service.ts
lilyydu: Ooh okay, I was thinking of doing so in case a user uploads a small photo that's hard to see but might not be a problem since it's on the Web

src/services/AmazonS3Service.ts
lilyydu: Ooh sorry, I meant imageType to jpeg, gif, png, etc- the ones you mentioned somewhere as a comment!

src/components/ImageSelector.tsx
ericrkuo: Awesome thanks!

src/services/AmazonS3Service.ts
ericrkuo: Oh, how did we configure the key to equal `promotionId.imageType`?

src/components/ImageSelector.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <>
+      {/* Displays the image selector if no image is selected */}
+      {!image.imageBinary && (
+        <input
+          type="file"
+          accept="image/jpeg, image/png, image/gif"
+          onChange={(event) => {
+            AmazonS3Service.onFileChange(event)
+              .then((image) => setImage(image))
+              .catch((err) => alert(err));
```
andrewzulaybar: Can you add types to the parameters of both these arrow functions? I think what might be good would be to have a `types/amazon.ts` file where you can place the Image interface and then just export it from there

src/components/ImageSelector.tsx
andrewzulaybar: Ohh can you also add the links to these tickets in the TODO comment here?

src/components/ImageSelector.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { ReactElement, useState } from 'react';
+
+import AmazonS3Service from '../services/AmazonS3Service';
+
+export default function ImageSelector(): ReactElement {
+  const [image, setImage] = useState({ imageBinary: '', imageType: '' });
```
andrewzulaybar: We can do:
```suggestion
  const [image, setImage] = useState<Image>({ imageBinary: '', imageType: '' });
```
so that setImage is typed i.e. it shouldn't let you use it without passing in a proper "Image" JSON object

src/services/AmazonS3Service.ts
andrewzulaybar: Oops didn't see this -- it would be good to use it in the ImageSelector component!

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+class AmazonS3Service {
+  onFileChange(e: React.ChangeEvent<HTMLInputElement> | DragEvent): Promise<Image> {
+    // Get the file from selection or drag and drop
+    const files =
+      (e as React.ChangeEvent<HTMLInputElement>)?.target?.files ||
+      (e as DragEvent)?.dataTransfer?.files ||
+      [];
+    if (!files.length) return Promise.resolve({ imageBinary: '', imageType: '' });
```
andrewzulaybar: Could you wrap the if-then block with curly braces?

```suggestion
    if (!files.length) {
      return Promise.resolve({ imageBinary: '', imageType: '' });
    }
```

Just to stay consistent with code style!

src/services/AmazonS3Service.ts
andrewzulaybar: I think we can keep the MAX_IMAGE_SIZE as a constant at the top of this file instead of an environment variable, and then we also won't have to cast it to a string

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,75 @@
...
+      reader.readAsDataURL(file);
+    });
+  }
+
+  uploadImage(image: Image, promotionId: string) {
+    // Get the presigned URL
+    axios
+      .get(process.env.REACT_APP_AWS_PRESIGNED_URL_ENDPOINT as string, {
+        params: {
+          promotionId: promotionId,
```
andrewzulaybar: Shorthand:
```suggestion
          promotionId,
```
Which is the same thing as you have here, just eliminates redundancy

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,75 @@
...
+          imageType: image.imageType,
+        },
+      })
+      .then((response: AxiosResponse) => {
+        // console.log('Presigned URL response: ', response)
```
andrewzulaybar: Do you still need this?

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,75 @@
...
+        // Upload the image
+        const binary = atob(image.imageBinary.split(',')[1]);
+        const array = [];
+        for (let i = 0; i < binary.length; i++) {
+          array.push(binary.charCodeAt(i));
+        }
+        const blobData = new Blob([new Uint8Array(array)], {
```
andrewzulaybar, line 57:
```suggestion
        const blobData = new Blob([new Uint8Array(binary.split("")], {
```
Does this also work? I think splitting with an empty string splits on every character

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,75 @@
...
+          type: 'image/' + image.imageType,
+        });
+        // console.log('Uploading image to: ', response.data.uploadURL)
+        const result = fetch(response.data.uploadURL, {
+          method: 'PUT',
+          body: blobData,
+        });
+        // console.log('Result: ', result)
```
andrewzulaybar: Do you need this too?

andrewzulaybar (changes requested):
Looks great! I just made some TS comments mainly :)

ericrkuo (approved):

src/services/AmazonS3Service.ts
richardyjtian: I was just thinking that because this is all client facing, we should try to restrict users from being allowed to change how large  the image size can be. But then again, the user could potentially just remove this line too, so it might not make a big difference.

src/components/ImageSelector.tsx
richardyjtian: Sounds good, I think types/image.ts might be more appropriate because it represents a generic image

src/services/AmazonS3Service.ts
richardyjtian: Nope, it's just for debugging. I've removed it

src/services/AmazonS3Service.ts
richardyjtian: This doesn't, because array is an array of character codes

src/services/AmazonS3Service.ts
richardyjtian: No, but we might need this later in case we want to display something if it fails.

src/services/AmazonS3Service.ts
andrewzulaybar: Ah sorry, just taking a look at this now! That makes sense, yeah we can keep it then :)

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,74 @@
...
+        // Upload the image
+        const binary = atob(image.imageBinary.split(',')[1]);
+        const array = [];
+        for (let i = 0; i < binary.length; i++) {
+          array.push(binary.charCodeAt(i));
+        }
+        const blobData = new Blob([new Uint8Array(array)], {
+          type: 'image/' + image.imageType,
+        });
+        const result = fetch(response.data.uploadURL, {
```
andrewzulaybar: For this, are you able to use axios instead of fetch? I was thinking because you used axios on line 43 we could just keep using that

src/services/AmazonS3Service.ts
```diff
@@ -0,0 +1,74 @@
+import axios, { AxiosError, AxiosResponse } from 'axios';
+
+import { Image } from '../types/image';
+
+const maxImageSize = 1000000;
```
andrewzulaybar: Can you add a comment or maybe change the variable name to include the units that this is in?

andrewzulaybar (approved):
Sorry for the late review, but this looks really great! Just two more comments to consider but everything else looks good :)

src/services/AmazonS3Service.ts
richardyjtian: I don't think axios allows uploading blob, so we might have to stick with fetch

andrewzulaybar: Oh interesting, sounds good then we can keep this!

PR #25, opened 2021-02-24
PP-28, PP-52: Implement save and delete buttons for promotion cards

andrewzulaybar:
For the promotion cards, this PR allows us to press the heart icon to like a promotion and the trash icon to delete the promotion (only for the promotions that the user has uploaded).

**Implementation details:**
- Added `DeleteModal` component, which prompts the user to confirm the promotion deletion
- `src/index.css` now has CSS variables which can be accessed throughout app: usage e.g. `color: var(--red)`
- Initial render of `/myAccount` page will fetch the logged in user's uploaded promotions
- Used `axios` to make BE requests, making processing API responses easier

**Notable changes included:**
- Added "@typescript-eslint/explicit-function-return-type" to eslint: to enforce us to type the return values of functions
- Instead of destructuring values in the props of a component, we leave the parameters of a component as `props: Props` where `Props` is defined as an interface with the typed properties. This makes it easier to scan through components because when a value is referenced, we know immediately whether it comes from `props` or not
- When passing in props to a component, we explicitly list them out instead of spreading out values (e.g. not `<PromotionCard {...promotion}>`). Even though it's more verbose, it's better to be explicit because of unnecessary re-renders etc.

**Demo:**
https://www.loom.com/share/290d4fe44fd24b73896c5fe3ec743cf6


src/components/PromotionList.tsx
```diff
@@ -29,6 +30,33 @@ export default function PromotionList({
...
+        return UserService.unsavePromotion(promotionId)
+          .then(() => {
+            promotion.liked = true;
+            setPromotions(promos);
+          })
+          .catch(() => null);
+      }
+    },
+    [promotions, setPromotions]
+  );
```
ericrkuo: How would we remember whether the promotion is `liked` if we refresh the page? Because it seems like `liked` isn't a property we have in the BE for a promotion, so it would default to false once we refresh the page right.

But it does make sense for the BE to provide the `liked` property though for a promotion...the BE would need to know which user is making request for GET promotions. Hmmm this will require some new behaviour in the BE, let me try and make a PR really quick to support this

ericrkuo: I'm thinking we might have to change these two routes to include the `id` of the user. Because now these endpoints need to have context of the user. But at the same time, I'm not a huge fan of making these major endpoints user specific. I think I'll need to research different ways of doing this

https://github.com/ubclaunchpad/PromoPal-backend/blob/master/src/main/route/PromotionRouter.ts#L12-L13

andrewzulaybar: @ericrkuo What do you think about just creating a duplicate endpoint for the first one like `GET /promotions/:userId` which returns the liked property? I'm not sure if we need the second one with the promotionId

ericrkuo: I think if we've authorized the request, we can use the user in the request headers. Then we don't need to make a new endpoint, we just modify the existing one to use the user based on the authorization

ericrkuo: At the same time, I think doing this is risky in terms of we need user auth in the FE to be complete first before we make the GET promotions endpoint authorized.

andrewzulaybar: Oh yeah that's a good point, forgot about that. And I guess that makes it easy because if a user isn't logged in then we'd just fetch all the promotions without joining

I think it's ok if we do it now actually since Yu merged his work in. Running the most up-to-date versions of the FE and the BE together is currently already broken because of user auth :(

ericrkuo: Oh it's broken at the moment? Hmm, probably next time we shouldn't merge in the auth without having it done in the FE too.

Hmm if the user isn't logged in, I think the whole endpoint wouldn't work because our authentication would fail at the middleware level. That's why I'm kind of thinking it's very complex to have our application work for both non-authenticated (users who did not sign up) and authenticated users.

I can work on implementing the changes in GET /promotions, but I think we should fail the request if the user is not authenticated and probably if we can, abandon the idea of having our app work without users signing up first

ericrkuo: Actually, I think we don't need to put user auth in the GET /promotions, just an optional parameter `userId` in the request params. But I still think we should consider avoiding making our application behave differently for non-auth vs auth users.

andrewzulaybar: Oh sorry maybe it's just the Uploaded Promotions but I think it's just the User endpoints because of the new authentication, but yeah you're right we should have waited until they were both complete

And yeah those are good points, it does also increase complexity on the FE to have to provide an unauthenticated view on top of the authenticated one. I think in terms of UX it probably will deter usage to make them sign up but for meeting our deadline it probably isn't the largest concern? So yeah for now let's abandon that idea

andrewzulaybar: Okay yeah that makes sense!

ericrkuo: Okie I think I can get this done by Wed night and I'll update you with changes you need to make in this PR when I create my PR, would that be okay?

andrewzulaybar: Oh yes that would be great! No rush if you're busy but let me know when you get to it :)

ericrkuo:
@andrewzulaybar made the necessary changes https://github.com/ubclaunchpad/PromoPal-backend/pull/126. rather than `liked`, I named it `isSavedByUser` to make it easier to understand. You will need to modify GET /promotions to take in a userId

src/components/promotion/PromotionCard.tsx
```diff
@@ -23,17 +23,45 @@ const styles: { [identifier: string]: CSSProperties } = {
...
+interface Props {
+  dateAdded: string;
+  description: string;
+  expirationDate: string;
+  id: string;
+  image: PromotionImageType;
+  name: string;
+  placeId: string;
+  restaurantName: string;
+  savedByUser: boolean;
```
ericrkuo: do we want to rename the prop to `isSavedByUser`? That way we make it consistent with the `Promotion` interface

ericrkuo (approved):

ericrkuo:
Feel free to merge in https://github.com/ubclaunchpad/PromoPal-backend/pull/126 when you're going to merge in this PR. I haven't got 2 approvals in the BE yet

PR #27, opened 2021-02-28
PP-26: Implement promotions tab

andrewzulaybar:
This PR implements the design and sets up the data fetching architecture for the `Promotions` tab in the Restaurant card. However, some work still needs to be incorporated to fetch the promotions dynamically. The implementation deviates from the original design a bit but I'm open to any suggestions and feedback!

**Implementation details:**
- Currently, just uses all the promotions but we want to only fetch the promotions for the currently open restaurant
- Created `time.ts` util file for functions related to time manipulation and formatting

**Photo:**
PromotionsTab


src/utils/time.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @param endTime - The end time to parse
+ */
+export function formatTime(startTime: string, endTime: string): string {
+  const formatHour = (startTime: string): string => {
+    const [hour, minute] = startTime.split(':');
+
+    const hourNum = parseInt(hour);
+    const minuteNum = parseInt(minute);
+
+    let formattedTime = `${hourNum === 12 ? hourNum : hourNum % 12}`;
```
richardyjtian: startTime and endTime are 24h right? So should this be `hourNum <= 12 ? hourNum : hourNum % 12` ?

src/types/promotion.ts
```diff
@@ -75,4 +75,5 @@ export interface FilterOptions {
   promotionType: Array<Promotion['promotionType']>;
 }
 
-type Day = 'Sunday' | 'Monday' | 'Tuesday' | 'Wednesday' | 'Thursday' | 'Friday' | 'Saturday';
+const DAY = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'] as const;
+export type Day = typeof DAY[number];
```
richardyjtian, line 79: This is really smart

richardyjtian (commented):
Great work Andrew! I can tell you put a ton of thought into how everything integrates together.

src/utils/time.ts
ericrkuo: This sounds like there is an npm library that can help us do this!

andrewzulaybar: @richardyjtian Oh yeah those should be the same I think? If `hourNum < 12` then this returns `hourNum % 12` which equals `hourNum` but thinking about it more I like your suggestion because it's clearer

@ericrkuo Yeah, I thought date-fns would be good for this and I played around with it but because `startTime` and `endTime` are strings and not Date objects, we would need to do more work to convert the strings to Dates first before we can pass it to other libraries. So it might make this function more complicated if we do

src/types/promotion.ts
ericrkuo: Oh wait do you mind explaining how this works?

src/utils/time.ts
ericrkuo: Ohh I see, makes sense sounds good. Awesome!

ericrkuo:
In my PR, I created a restaurant router and I think I can implement an endpoint called GET restaurants/:id/promotions, which will get all the promotions for a restaurant. Then we could use this to dynamically load in the data for this part. What do you think about this idea? If it's good, could we create a ticket in JIRA to track this in the BE and also link the ticket to the FE ticket?
https://github.com/ubclaunchpad/PromoPal-backend/pull/123

src/types/promotion.ts
andrewzulaybar: Yeah! I got this from Stack Overflow: https://stackoverflow.com/questions/44480644/string-union-to-string-array

```
const DAY = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'] as const;
```

This is a const assertion, more info [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) but this is the important line: `array literals become readonly tuples`. So `DAY` is now a readonly tuple.

```
export type Day = typeof DAY[number];
```
And with `DAY[number]` here, since `DAY` is immutable and it knows that it can only be indexed with numbers 0-6, `typeof DAY[0]` is the same as `'Sunday'`, `typeof DAY[1]` is the same as `'Monday'` etc. so `Day` is now a discriminating union that contains all these types so this is the same as:
```
export type Day = 'Sunday' | 'Monday' | 'Tuesday' | 'Wednesday' | 'Thursday' | 'Friday' | 'Saturday';
```

andrewzulaybar:
> In my PR, I created a restaurant router and I think I can implement an endpoint called GET restaurants/:id/promotions, which will get all the promotions for a restaurant. Then we could use this to dynamically load in the data for this part. What do you think about this idea? If it's good, could we create a ticket in JIRA to track this in the BE and also link the ticket to the FE ticket?
> [ubclaunchpad/PromoPal-backend#123](https://github.com/ubclaunchpad/PromoPal-backend/pull/123)

That would actually be super helpful thanks! I can make some changes here to use that :)
https://promopal.atlassian.net/browse/PP-88

src/types/promotion.ts
ericrkuo: Woahh super cool!! Thanks for the explanation

ericrkuo:
> > In my PR, I created a restaurant router and I think I can implement an endpoint called GET restaurants/:id/promotions, which will get all the promotions for a restaurant. Then we could use this to dynamically load in the data for this part. What do you think about this idea? If it's good, could we create a ticket in JIRA to track this in the BE and also link the ticket to the FE ticket?
> > [ubclaunchpad/PromoPal-backend#123](https://github.com/ubclaunchpad/PromoPal-backend/pull/123)
> 
> That would actually be super helpful thanks! I can make some changes here to use that :)
> https://promopal.atlassian.net/browse/PP-88

Cool, so we will merge this in, do the BE, then come back to here and tackle connecting the BE?

ericrkuo (approved):

PR #28, opened 2021-02-28
PP-48: Connect account details form to BE

andrewzulaybar:
This PR connects Richard's work with the `AccountDetails` component to the BE. Still have to look into how Yu's work with Firebase will affect the structure here.

**Implementation details:**
- Adds validation rules to input values so that we can prevent form submission with invalid values
- Added methods to `UserService` which make the associated requests on the BE

**Notes:**
- We still need to handle errors when we receive a bad response from the BE after updating account details: https://promopal.atlassian.net/browse/PP-80


src/services/UserService.ts
```diff
@@ -1,19 +1,64 @@
...
-  public async getUploadedPromotions(userId: string): Promise<Promotion[]> {
-    const url = Routes.USERS.UPLOADED_PROMOTIONS(userId);
-    return fetch(url)
-      .then((response: Response) => response.json())
-      .then((response: UploadedPromotionsResponse) => {
-        if (isError<UploadedPromotionsResponse>(response)) {
-          return Promise.reject(response);
+  private userId: string;
+
+  public constructor() {
```
richardyjtian: Would we save userId as a cookie after the user authenticates, and have this constructor get the userId from the cookie every time the web app is reopened?

ericrkuo: @yunishimura0716 do you mind answering this question?

yunishimura0716: I would say yes.
I'm now thinking about the best way to store user info (`userId` in this case), Local storage or cookie or other??.

andrewzulaybar: I think local storage is probably a good and common solution! We can persist it to local storage for a specific amount of time and then when it's invalidated, just log out the user

yunishimura0716: @richardyjtian I realized that it is impossible to get `userId` for current flow of `signIn` since it just authorize by firebase server, not our db. I will change the implementation of BE to be able to request GET `users/:id` where `id` is user id of firebase not our db. So, we don't have to store `userId` of our db since we can get `id` of firebase by using property of `firebase currentUser`. This means, when FE request GET `users/:id`, it just set `id` of firebase.

richardyjtian (approved):
This looks good! I'll integrate it with firebase when it's merged

PR #29, opened 2021-03-01
PP-31, PP-40: Create log-in component, create registration component

engineereng:
Adds a login page and registration page under the /account route, which replaces /login. Clicking the "Register here!" button moves the user to the Registration page, and clicking the back button from the Registration page moves the user back.

login-screen
registration-screen

The inputs check whether the user has input all of the necessary fields. Additionally, the Registration page validates the format of the email field, and the password field requires a password that is >=9 characters long.

Currently, there is no way to check whether the user is logged in. When they are, the Login button will instead be a "My Account" button that redirects to the My Account screen.

Next step is to integrate with Firebase.


src/components/account/LoginCard.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export default function LoginCard(props: Props): ReactElement {
+  const onFinish = () => {
+    alert('Finish');
+    //console.log('Success:', values);
+  };
+
+  const onFinishFailed = () => {
+    //console.log('Failed:', errorInfo);
+  };
```
ericrkuo: Could we link the tickets as todo comments here? If the tickets don't exist in the `User Authentication` epic in JIRA, would you mind creating them? That way we can keep track of what's left in the FE.

ericrkuo (approved):
Beautiful work @engineereng

src/components/account/LoginCard.tsx
engineereng: Sure!

richardyjtian (approved):
Looks great to me! I'll begin on the FE user auth when this is merged

src/components/account/LoginCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+  },
+};
+
+const layout = {
+  labelCol: { span: 4 },
+  wrapperCol: { span: 8 },
+};
+const tailLayout = {
+  wrapperCol: { offset: 4, span: 8 },
+};
```
andrewzulaybar, line 23: Minor thing, can you add a line between these variable declarations? 😛

src/components/account/LoginCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+        </Form.Item>
+
+        <Form.Item {...tailLayout}>
+          <Button className="button" style={styles.button} htmlType="submit">
+            Login
+          </Button>
+        </Form.Item>
+      </Form>
+
+      <p className="inputLabel">Don't have an account yet?</p>
```
andrewzulaybar: For consistency, let's use kebab-case for class names! So this would be `input-label` instead

src/components/account/LoginCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <Button
+        size="large"
+        shape="round"
+        onClick={props.onClickRegister}
+        className="button"
+        style={styles.button}
+      >
```
andrewzulaybar:
```suggestion
      <Button
        size="large"
        shape="round"
        type="submit"
        onClick={props.onClickRegister}
        className="button"
        style={styles.button}
      >
```

Just add `htmlType="submit"` and because this button will be our trigger to submit the form, we want to move it up within the form (i.e. before the </Form> on line 76) and wrap it in a `Form.Item`!

src/components/account/RegisterCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+      >
+        <Input placeholder="Username" />
+      </Form.Item>
+      <Form.Item
+        style={styles.inputWrapper}
+        name="password"
+        label="Password"
+        rules={[
+          { required: true, message: 'A password is required.' },
+          { min: 9, message: 'Passwords must be at least 9 characters.' },
```
andrewzulaybar, line 96: Nice! Good call on adding these additional rules here

src/components/account/RegisterCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+        ]}
+        hasFeedback
```
andrewzulaybar:
```suggestion
        hasFeedback={true}
```
Just to be a bit more explicit here

src/components/account/RegisterCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+          ({ getFieldValue }) => ({
+            validator(_, value) {
+              if (!value || getFieldValue('password') === value) {
+                return Promise.resolve();
+              }
+              return Promise.reject(new Error('The two passwords that you entered do not match!'));
+            },
+          }),
+        ]}
+        hasFeedback
```
andrewzulaybar:
```suggestion
        hasFeedback={true}
```

src/components/account/RegisterCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+      >
+        <Input.Password placeholder="Confirm Password" />
+      </Form.Item>
+      <Form.Item>
+        <Button className="button" style={styles.button} htmlType="submit">
+          Register
+        </Button>
+      </Form.Item>
```
andrewzulaybar, line 128: For better readability, can you add an empty line between each `Form.Item`?

src/components/account/ResetPassword.tsx
```diff
@@ -32,7 +26,7 @@ export default function ResetPassword(): ReactElement {
 
   const InputPasswordWrapper = ({ label }: { label: string }): ReactElement => (
     <div style={styles.inputWrapper}>
-      <p style={styles.inputLabel}>{label}</p>
+      <p className="inputLabel">{label}</p>
```
andrewzulaybar:
```suggestion
      <p className="input-label">{label}</p>
```

src/index.css
```diff
@@ -17,3 +17,14 @@ code {
...
 }
+
+.ant-btn.button {
+  background-color: #ffc529;
+  float: right;
+  font-weight: bold;
+  margin-top: 10;
+}
+
+.inputLabel {
```
andrewzulaybar, line 28:
```suggestion
.input-label {
```

src/screens/Login.tsx
```diff
@@ -0,0 +1,31 @@
...
+  container: {
+    backgroundColor: 'white',
+    borderRadius: 20,
+    boxShadow: '0 4px 4px 0 #40333333',
+    color: '#333',
+    margin: 30,
+    overflow: 'auto',
+    padding: 30,
+  },
+};
```
andrewzulaybar, line 19: Add another empty line after this

andrewzulaybar (changes requested):
@engineereng This looks super great!!! 🙌

Just added a bunch of minor comments here mainly for code style cause I haven't gone through it with you yet :)

src/components/account/LoginCard.tsx
richardyjtian: I'm not sure if we want the "Register" button to submit the "Login" form. The button can move above, but it should redirect to the registration page

andrewzulaybar: Ahh you're right, thanks for catching this!

andrewzulaybar (approved):
Looks good! We can hold off on my suggestions until a later time because they're mostly stylistic and it's currently blocking @richardyjtian. 

Merge whenever :)

PR #30, opened 2021-03-07
PP-78 Refactor Google Places Behaviour

ericrkuo:
https://promopal.atlassian.net/browse/PP-78
Please see https://github.com/ubclaunchpad/PromoPal-backend/pull/123/ for more information on BE changes

- We no longer need to interact with google places directly from the FE because this is now happening in the BE
- A `Promotion` has a `Restaurant` which has a `placeId, lat, lon`
- FE needs to call the new BE endpoint `GET /promotions/:id/restaurantDetails/:placeId`


package.json
```diff
@@ -30,6 +30,7 @@
     "@ant-design/icons": "^4.2.2",
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
+    "@googlemaps/google-maps-services-js": "3.1.16",
```
ericrkuo, line 33: This library is not recommended to be used for client-side, but I am only using it for it's interfaces

src/components/MapContainer.tsx
```diff
@@ -25,22 +25,22 @@ function MapContainer({
...
 
       if (map) {
-        promotionsState.data.forEach(({ lat, lon, placeId }) => {
-          const position = { lat, lng: lon };
+        promotionsState.data.forEach(({ restaurant }) => {
+          const position = { lat: restaurant.lat, lng: restaurant.lon };
           const marker = new google.maps.Marker({ position, map });
-          marker.addListener('click', () => onClickHandler(placeId));
+          marker.addListener('click', () => onClickHandler(restaurant.id));
         });
```
ericrkuo, line 44: I was thinking rather than using `placeId`, we use the `id` of the restaurant. This is because each restaurant is associated with a `placeId`.

src/components/promotion/PromotionDetails.tsx
```diff
@@ -98,9 +97,6 @@ export default function PromotionDetails({
           <HeartOutlined style={styles.heart} />
         </Col>
       </Row>
-      <Row>
-        <Title style={styles.restaurantName}>{restaurantName}</Title>
-      </Row>
```
ericrkuo, line 90: We are no longer storing the user inputted restaurant names in the BE. This should be in the restaurant card ideally

src/components/restaurant/RestaurantCard.tsx
```diff
@@ -35,17 +35,25 @@ export default function RestaurantCard(restaurant: Restaurant): ReactElement {
...
+          // cuisine={
+          //   'Italian'
+          //   // todo: this should use a promotions cuisine, google place details doesn't have this
+          //   // restaurant.cuisine
+          // }
+          // distance={
+          //   1500
+          //   // // todo: where is this coming from?
+          //   // restaurant.distance
+          // }
```
ericrkuo: @andrewzulaybar I did not see where we can get this information from the place details. Do you mind explaining how we are getting these values?

src/components/restaurant/card/Body.tsx
```diff
@@ -66,9 +54,15 @@ export default function Body({
     <Row>
       <Tabs style={styles.tabs} defaultActiveKey="1" size="small">
         <TabPane tab="Info" key="1" style={styles.tabContent}>
-          <Section title="Address" details={address} />
-          <Section title="Hours" details={formatHours(openingHours)} />
-          <Section title="Phone" details={formatPhoneNumber(phoneNumber)} />
+          <Section title="Address" details={formatted_address ?? 'No address found'} />
+          <Section
+            title="Hours"
+            details={opening_hours?.weekday_text.join('\r\n') ?? 'No available hours found'}
```
ericrkuo, line 56: This now looks like
```
        Monday: 11:00 AM – 10:00 PM
        Tuesday: 11:00 AM – 10:00 PM
        Wednesday: 11:00 AM – 10:00 PM
        Thursday: 11:00 AM – 10:00 PM
        Friday: 11:00 AM – 10:00 PM
        Saturday: 11:00 AM – 10:00 PM
        Sunday: 11:00 AM – 10:00 PM
```

src/services/GooglePlacesService.ts
```diff
@@ -1,183 +1,48 @@
...
 import axios, { AxiosError, AxiosResponse } from 'axios';
 
-import { RestaurantDetails } from '../types/restaurant';
-import { RestaurantInfo } from '../types/RestaurantInfo';
+import Routes from '../utils/routes';
 
-/* eslint-disable  @typescript-eslint/no-explicit-any */
 class GooglePlacesService {
-  // maintains a mapping of restaurant placeIDs with associated restaurant details (see types)
-  private currRestaurants: Map<string, RestaurantDetails>;
```
ericrkuo, line 9: Now this class is much simpler, leaving the responsibility of handling NOT_FOUND's to the BE, using interfaces etc

src/types/promotion.ts
```diff
@@ -68,6 +62,13 @@ export interface Schedule {
   isRecurring: boolean;
 }
 
+export interface Restaurant {
+  id: string;
+  placeId: string;
+  lat: number;
+  lon: number;
+}
```
ericrkuo, line 70: New interface `Restaurant`. Same as https://github.com/ubclaunchpad/PromoPal-backend/blob/PP-77-Refactor-Google-Place-Behaviour/src/main/entity/Restaurant.ts

src/components/MapContainer.tsx
ericrkuo: Lol my random latitude and longitude in the BE has made the restaurants spread out throughout the world

src/services/GooglePlacesService.ts
```diff
@@ -1,183 +1,48 @@
...
-      });
+  /**
+   * Gets photos for a restaurant
+   * @param restaurantId the id of the restaurant
+   * */
+  async getRestaurantPhoto(restaurantId: string): Promise<PlacePhoto[]> {
+    const place = await this.getRestaurantDetails(restaurantId);
+    return place.photos ?? [];
   }
 }
```
ericrkuo: We should be unblocked now for https://promopal.atlassian.net/browse/PP-27.

ericrkuo:
@andrewzulaybar if you wouldn't mind, it would be great if you could double-check on your end that the state for restaurants are being propagated correctly in terms of https://github.com/ubclaunchpad/PromoPal-frontend/blob/6efe2bb89c574e9a7def72932d2715870bb58868/src/contexts/RestaurantCardContext/types.ts#L37-L49

If you could, thanks a million! You would need your BE branch to be `PP-77-Refactor-Google-Place-Behaviour` and don't forget to run yarn install for both!

src/components/promotion/PromotionDetails.tsx
andrewzulaybar: Oh I see... I think we could replace this with a call to get the name via the `use-places-autocomplete` library once the upload promotions PR is merged in

src/components/restaurant/RestaurantCard.tsx
andrewzulaybar: Oh yes cuisine should be from the promotion object and distance I think we still have to calculate on the FE

Just added this for distance:
https://promopal.atlassian.net/browse/PP-89

src/components/restaurant/RestaurantCard.tsx
```diff
@@ -1,7 +1,7 @@
+import { Place } from '@googlemaps/google-maps-services-js';
```
andrewzulaybar, line 1: I'm wondering, instead of this library are we able to make use of https://www.npmjs.com/package/@types/googlemaps since it's already installed? I think there's a `PlaceResult` interface here that looks similar to the one you're using in this file

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/googlemaps/reference/places-service.d.ts

src/components/restaurant/RestaurantCard.tsx
```diff
@@ -35,17 +35,25 @@ export default function RestaurantCard(restaurant: Restaurant): ReactElement {
...
           rating={restaurant.rating}
           website={restaurant.website}
         />
         <Body
-          address={restaurant.address}
-          openingHours={restaurant.openingHours}
-          phoneNumber={restaurant.phoneNumber}
+          formatted_address={restaurant.formatted_address}
+          opening_hours={restaurant.opening_hours}
+          formatted_phone_number={restaurant.formatted_phone_number}
```
andrewzulaybar, line 57: For consistency, could we use camelCase here for the props? e.g. formattedAddress

src/components/restaurant/card/Body.tsx
```diff
@@ -37,21 +36,10 @@ function Section({ title, details }: { title: string; details: string }): ReactE
...
-      Wed: ${hours.wednesday}
-      Thu: ${hours.thursday}
-      Fri: ${hours.friday}
-      Sat: ${hours.saturday}
-    `;
-  }
-
+  formatted_address,
+  opening_hours,
+  formatted_phone_number,
```
andrewzulaybar, line 44: camelCase here

src/components/restaurant/card/Header.tsx
```diff
@@ -50,16 +50,13 @@ const styles: { [identifier: string]: CSSProperties } = {
 };
 
 export default function Header({
-  cuisine,
-  distance,
+  // cuisine,
+  // distance,
   name,
-  priceLevel,
+  price_level,
```
andrewzulaybar, line 56: camelCase here

src/reducers/RestaurantCardReducer.ts
```diff
@@ -15,10 +16,10 @@ export function restaurantCardReducer(state: State, { type, payload }: DispatchP
      * Shows the restaurant card.
      */
     case DispatchAction.SHOW_CARD: {
-      const { placeId, restaurant } = payload as { placeId: string; restaurant: Restaurant };
+      const { restaurantId, restaurant } = payload as { restaurantId: string; restaurant: Place };
       nextState = {
         ...nextState,
-        placeId,
+        restaurantId: restaurantId,
```
andrewzulaybar:
```suggestion
        restaurantId,
```

src/services/GooglePlacesService.ts
```diff
@@ -1,183 +1,48 @@
...
-    const dataURI: string = encodeURIComponent(restaurantName);
-    let responseData: any[];
+  /**
+   * Gets all restaurant details for certain restaurant
+   * @param restaurantId the id of the restaurant
+   */
+  getRestaurantDetails(restaurantId: string): Promise<Place> {
+    // if mapping already contains this restaurantId, return the place
+    const existingPlace = this.currRestaurants.get(restaurantId);
+    if (existingPlace) return Promise.resolve(existingPlace);
```
andrewzulaybar:
```suggestion
    if (existingPlace) {
      return Promise.resolve(existingPlace);
    }
```

src/services/GooglePlacesService.ts
```diff
@@ -1,183 +1,48 @@
...
-      })
-      .catch((err: AxiosError) => {
-        return Promise.reject(err);
-      });
+  /**
+   * Gets photos for a restaurant
+   * @param restaurantId the id of the restaurant
+   * */
+  async getRestaurantPhoto(restaurantId: string): Promise<PlacePhoto[]> {
+    const place = await this.getRestaurantDetails(restaurantId);
```
andrewzulaybar: Do we want to wrap this in a try/catch or use then/catch blocks in case it fails?

src/services/PromotionService.ts
```diff
@@ -1,6 +1,12 @@
+import {
+  AddressType,
+  GeocodingAddressComponentType,
+  LocationType,
+  Place,
```
andrewzulaybar, line 5: These types should have matching counterparts in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0b5bfba2994c91a099cd5bcfd984f6c4c39228e5/types/google__maps/index.d.ts

andrewzulaybar (changes requested):
@ericrkuo Great work, and thanks for looking into this!

I made some comments, mainly just wondering if you'd be able to look into the type library I mentioned and whether they already have the associated types.

src/components/restaurant/RestaurantCard.tsx
ericrkuo: Oh okay this might be difficult since many promotions can belong to the same restaurant, and I can't confidently say the cuisine will be the same for all promotions of the same restaurant. (for example fusion restaurant chinese + mexican, we have a chinese promo and a mexican promo for same restaurant)

src/components/promotion/PromotionDetails.tsx
ericrkuo: Oh how come we want to call autocomplete here?

src/components/restaurant/RestaurantCard.tsx
ericrkuo: ah yes, we could use it, but BE is currently sending data in the format of the `Place` interface. `PlaceResult` seems like the same, but I'm not sure if I wanna take the risk that they'll remain the same in the future.

The reason why I chose this library in the BE is cause of the following pros where old=@types/googlemaps and new=@googlemaps/google-maps-service-js
image

src/components/promotion/PromotionDetails.tsx
andrewzulaybar: Oh because we're not storing the restaurant name in the BE anymore, but I think it'd probably be good to still display the restaurant name here because it prevents users from having to open up the restaurant card to see the name

src/components/restaurant/RestaurantCard.tsx
andrewzulaybar: Ahh I see what you mean... Okay I think this complicates things lol I've created a ticket to think about it more

But I think we can probably implement some algorithm to show maybe the top 2/3 cuisine types for the restaurant based on the promotions or something like that

https://promopal.atlassian.net/browse/PP-91

src/components/restaurant/RestaurantCard.tsx
andrewzulaybar: Ohhh didn't know this, that's good to know! Yeah in that case it's probably better to stick with @googlemaps/google-maps-service-js

andrewzulaybar:
@ericrkuo This is what I see on my end! Only issue I see is the phone number. And for the restaurant card context seems to be working ok too :)

Screen Shot 2021-03-19 at 1 26 17 PM

src/components/restaurant/card/Body.tsx
ericrkuo: I'm not too sure I can do all these camel Case's because they're coming from the `Place` interface

src/components/promotion/PromotionDetails.tsx
ericrkuo: I see, yeah I agree it's important to add the restaurant name here, but this might be difficult to do since when we are loading all the promotions, we'd need to get all the restaurant names for all promotions

ericrkuo:
> @ericrkuo This is what I see on my end! Only issue I see is the phone number. And for the restaurant card context seems to be working ok too :)
> 
> Screen Shot 2021-03-19 at 1 26 17 PM

Oops okie I fixed phone number, turns out it's already formatted so no need to format again

src/components/promotion/PromotionDetails.tsx
ericrkuo: I think for now, I can leave it uncommented and create a ticket once we have the information available. Because this would mean when we call GET /promotions, we'd need to call GET restuarantDetails for each promotion

src/components/restaurant/card/Body.tsx
andrewzulaybar: Actually I think I might have actually restructured the Prop types in my other PRs to match a new style so no need to do this anymore!

andrewzulaybar (approved):
Great work!!! 🎉  

Excited to get this in

PR #26, opened 2021-02-28
PP-83: FE - Add AutoComplete to Creating a Promotion page

ericrkuo:
https://promopal.atlassian.net/browse/PP-83

This PR is just research for autocomplete feature since we do not have an upload promotions page yet. Once we start on the upload promotions page, we can use the work done here already.

This PR confirms that the autocomplete functionality will work for our project. We can now get the `placeId` and `restuarantAddress`, so we no longer need to store `restuarantName` and `restaurantAddress` in the BE

Feel free to checkout this branch and try it out!

Demo:
AutocompleteGif

PR #8, opened 2021-01-31
[WIP]: Implement skeleton screen

andrewzulaybar:
## Description

_Currently a WIP until designs and workflows have been finalized_

PR #31, opened 2021-03-07
Add upload promotion form and Less.js

andrewzulaybar:
### ⚠️ Large PR, most of the line changes in this PR are from `yarn.lock` ⚠️

**Description:**
- Builds off of PR #8 but applies hi-fi styling
- Tried working with the library suggested in PR #26, but it ended up being hard to integrate with Ant Design and styling :( Instead, I found this library which actually ended up working well with our app: [use-places-autocomplete](https://www.npmjs.com/package/use-places-autocomplete)
- Adds [craco](https://github.com/gsoft-inc/craco) - Create React App Configuration Override, which allows us to make modifications to the build without having to eject the React app (i.e. less configuration files to deal with and manage)
- Adds [Less.js](http://lesscss.org/) - a CSS pre-processor that allows us to (a) customize the Ant Design styles and (b) write cleaner CSS files

Before Less, we would have had to find the Ant Design classes for each element we wanted to change the styles for:

With Less, we are able to leverage a lot of techniques to make CSS cleaner including variables and nesting classes.

**Changes:**
- The calls to fetching enums that were originally on the Home page were moved into the EnumService. This lets us only call these endpoints once (when the EnumService is loaded up) so that we can prevent unnecessary API requests.
- The location service I've separated into two base classes: `GeolocationPosition` and `LatLng`. Each base class just serves the functions and properties provided in the `LocationService` abstract class but returns the requested info in their respective types.
- `variables.less` should now be used to define global Less variables that we can use throughout component styling

**Todo:**
- handle errors in POST /promotion: https://promopal.atlassian.net/browse/PP-87
- add support for this when we handle discount values: https://promopal.atlassian.net/browse/PP-86
- calculate the `isRecurring` property for promotions
- need to check that form submission works as expected once https://github.com/ubclaunchpad/PromoPal-backend/pull/123 is merged in

**Demo:**
https://www.loom.com/share/450f046d28a943d080d8f0f881ffe258


src/components/upload-promotion/UploadPromotionForm.less
```diff
@@ -0,0 +1,38 @@
+.upload-promotion-form {
```
richardyjtian, line 1: So less allows you to define the scope of the ant properties to just the upload-promotion-form right? This is pretty useful

src/services/LocationService.ts
```diff
@@ -36,18 +53,51 @@ class LocationService {
...
       } else {
         resolve(this.defaultLocation);
       }
     });
   }
 }
 
-export default new LocationService();
+export default {
+  GeolocationPosition: new GeolocationPositionService(),
```
richardyjtian: Do we have two different location services in case one of the two doesn't work?

src/index.less
```diff
@@ -1,16 +1,19 @@
-@import url('https://fonts.googleapis.com/css2?family=ABeeZee&display=swap');
 @import url('https://fonts.googleapis.com/css2?family=Roboto&display=swap');
```
richardyjtian, line 1: Should we maybe just stick with one font throughout our app? I saw that Roboto is still used in theme.less

richardyjtian (approved):

src/components/upload-promotion/UploadPromotionForm.less
andrewzulaybar: Kind of! The scope of the ant rules defined in this file is already restricted to the files that import this file. But here Less allows us to do this "nesting" of CSS rules which just makes it easier to read for developers since it's in a hierarchy kind of format 😛

src/services/LocationService.ts
andrewzulaybar: Ahh the two location services are mainly to provide location info in different types, namely `GeolocationPosition` and `LatLng`. Some places we need `LatLng` objects (I think anything to do with Google APIs) whereas everywhere else we don't really need to convert the GeolocationPosition to LatLng types!

src/index.less
andrewzulaybar: Yeah I think we should -- as part of clean up in the next week or so I think we need to do an audit of styling on the frontend to make sure everything's consistent!

https://promopal.atlassian.net/browse/PP-100

PR #32, opened 2021-03-15
PP-11: Delegate ability to sort promotions to BE

andrewzulaybar:
**Description:**
Building off of https://github.com/ubclaunchpad/PromoPal-frontend/pull/20, this PR mainly serves to support sending the sorting request to the BE. Aside from that, there's a little bit of cleanup here.

**Related PRs:**
- https://github.com/ubclaunchpad/PromoPal-backend/pull/127

**Related tickets:**
https://promopal.atlassian.net/browse/PP-11


src/App.tsx
```diff
@@ -30,7 +30,7 @@ function App(): ReactElement {
 
   /**
    * Loads dependencies on initial render.
-   * Displays a spinner for the minimum of either: (a) the duration of the loading period or (b) 2 seconds.
+   * Displays a spinner for the maximum of either: (a) the duration of the loading period or (b) ~2 seconds.
```
richardyjtian, line 34: Should we maybe display a message telling the user to reload if the page is stuck somehow (load duration > some time)?

src/screens/Home.tsx
```diff
@@ -107,9 +108,9 @@ export default function Home(): ReactElement {
           action: () =>
             dispatch({
               type: DispatchAction.SORT,
-              payload: { sort: Sort.Rating },
```
richardyjtian, line 88: This is because we don't have an upvote/downvote functionality right?

src/types/promotion.ts
```diff
@@ -1,10 +1,10 @@
 import { RestaurantDetails } from './restaurant';
 
 export enum Sort {
```
richardyjtian, line 3: When is it best to use enum vs interface?

richardyjtian (approved):

src/App.tsx
andrewzulaybar: Oh yeah, good point. Not sure what we would display but maybe a small picture or something to go with it.

I added a ticket here to tackle it: https://promopal.atlassian.net/browse/PP-99

src/screens/Home.tsx
andrewzulaybar: Before we planned to do rating which is a property we could only get through the google places API -- and so to sort by rating we need to make an API call for each promotion which would be costly so I think we decided to do this instead

src/types/promotion.ts
andrewzulaybar: Yeah! So an interface is basically a JSON object, and it allows you to name the key and the type of the value:
e.g.
```typescript
interface SomeObject {
  aString: string;
  someOtherProperty: boolean;
  ...
}

// usage
const foo: SomeObject = {
  aString: "any string can go here",
  someOtherProperty: true,
 ... // other properties...
}
```

But an enum on the other hand only allows you take on ONE value at a time. So in this example:
```typescript
enum Sort {
  Default = 'Distance',
  Distance = 'Distance',
  Popularity = 'Popularity',
  Recency = 'Recency',
}

const option: Sort = Sort.Default; // usage
```
So `option` here can only either be `Sort.Default`, or `Sort.Distance`, or `Sort.Popularity`, or `Sort.Recency`. It can't take on more than one value at a time!

andrewzulaybar: Lemme know if that makes sense!

richardyjtian: Oooo I never knew this, thanks so much for explaining!

PR #33, opened 2021-03-20
PP-27: Implement restaurant photos

andrewzulaybar:
**Description:**
Implemented the display of restaurant photos in the restaurant card component. Builds off of the work that Eric's done in #30

**Related tickets:**
- https://promopal.atlassian.net/browse/PP-27


engineereng (approved):
Looks great to me!

src/services/GooglePlacesService.ts
```diff
@@ -43,6 +43,23 @@ class GooglePlacesService {
...
+   *
+   * @param photo the restaurant photo
+   */
+  getRestaurantPhotoUrl({ photo_reference }: PlacePhoto): string {
+    const baseUrl = 'https://maps.googleapis.com/maps/api/place/photo?';
+    const queryParams = new URLSearchParams({
+      maxheight: '1600',
+      maxwidth: '1600',
+      photoreference: photo_reference,
+      key: process.env.REACT_APP_GOOGLE_PLACES_API_KEY as string,
```
ericrkuo: I remember reading that we should use a separate google places API key that can only be used for image display purposes, (can't be used for place details or any other stuff). Because otherwise, this API key is accessible by any user, what do you think?

ericrkuo (approved):
Sorry I am very slow with my other PR at the moment, will try to get to that

src/services/GooglePlacesService.ts
andrewzulaybar: Ahh yeah, good point I forgot. Thanks for bringing this up!

ericrkuo (approved):

PR #36, opened 2021-03-21
PP-29: Integrate firebase auth services

richardyjtian:
Firebase auth is mostly integrated with the frontend following the design on https://docs.google.com/document/d/1RdtbGEHxUexCqKS7_DNWQT_68gSe7yL6uEvgYlgNCjs/edit. Services include account registration, login, sign out, password reset from email, and password change.

Account registration flow still needs to be completed in the backend in RegistrationService.tsx after a user is successfully registered in firebase (PP-36). Additional services to be added in the future include account deletion, and email change.

Resolves PP-35, PP-37, PP-41, PP-42, PP-43, PP-44, PP-45


src/components/account/LoginCard.tsx
```diff
@@ -28,24 +31,32 @@ interface Props {
...
-    alert('Finish');
-    //console.log('Success:', values);
-    // TODO https://promopal.atlassian.net/jira/software/projects/PP/boards/1?selectedIssue=PP-42
-    // TODO https://promopal.atlassian.net/jira/software/projects/PP/boards/1?selectedIssue=PP-43
+  const firebase = useFirebase();
+  const history = useHistory();
+
+  const onFinish = (data: { email: string; password: string; staySignedIn: boolean }) => {
+    firebase
+      .doSignInWithEmailAndPassword(data.email, data.password, data.staySignedIn)
```
yunishimura0716: This is kind of reminder. As we discussed before, the user have to request `/users/firebase/:firebaseId` after the user succeed to sign in by firebase.

src/contexts/AuthUserContext/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function AuthUserProvider({
+  children,
+}: {
+  children: ReactElement | ReactElement[];
+}): ReactElement {
+  const [authUser, setAuthUser] = useState<AuthUser | null>(null);
+  const firebase = useFirebase();
+
+  useEffect(() => {
```
yunishimura0716, line 27: I just want to ask you about `onAuthStateChanged`. Is this called every time the user `signIn` or `signUp` ?

yunishimura0716 (commented):
Great work, Richard! Thank you so much! The implementation makes sense to me. Some codes for auth flow with Backend is not implemented yet. If you have question for that, let me know.

src/contexts/AuthUserContext/index.tsx
richardyjtian: onAuthStateChanged is called every time the user signs in, but the user will automatically be signed in after signing up for an account too

src/services/RegistrationService.ts
```diff
@@ -0,0 +1,19 @@
+import Firebase from '../services/FirebaseService';
+import { RegistrationData } from '../types/user';
+
+class RegistrationService {
+  register(firebase: Firebase, data: RegistrationData) {
+    return firebase
+      .doCreateUserWithEmailAndPassword(data.email, data.password)
+      .then(() => {
+        // TODO: Register user in backend
+        // https://promopal.atlassian.net/browse/PP-36
```
ericrkuo: I think PP-36 has already been implemented, could we connect this to the BE?

src/contexts/AuthUserContext/index.tsx
yunishimura0716: Sounds good!

src/services/FirebaseService.ts
```diff
@@ -0,0 +1,75 @@
...
+  apiKey: process.env.REACT_APP_FIREBASE_API_KEY,
+  authDomain: process.env.REACT_APP_FIREBASE_AUTH_DOMAIN,
+  projectId: process.env.REACT_APP_FIREBASE_PROJECT_ID,
+  storageBucket: process.env.REACT_APP_FIREBASE_STORAGE_BUCKET,
+  messagingSenderId: process.env.REACT_APP_FIREBASE_MESSAGING_SENDER_ID,
+  appId: process.env.REACT_APP_FIREBASE_APP_ID,
+};
+
+class Firebase {
+  auth: firebase.auth.Auth;
```
andrewzulaybar: Can we make this private?

src/services/FirebaseService.ts
```diff
@@ -0,0 +1,75 @@
...
+      firebase.initializeApp(config);
+    } else {
+      firebase.app(); // if already initialized, use that one
+    }
+    this.auth = firebase.auth();
+  }
+
+  // *** Auth API ***
+  // Creates a user and signs the user in
+  doCreateUserWithEmailAndPassword = (email: string, password: string) => {
```
andrewzulaybar: Minor thing I forgot to mention last time, let's use regular function declarations instead of arrow functions for functions inside classes so:
```suggestion
  doCreateUserWithEmailAndPassword(email: string, password: string) {
```

andrewzulaybar: Also, just merged in a change to the linter that forces us to add types to return values of functions so you'll probably have to add that!

andrewzulaybar (approved):
This is really great, excited to get this in! I'll work on handling the error propagation this week. Just added a few comments.

Last few things we'll probably have to do is restrict access for certain pages if a user is not logged in, which we can work out this week. If there are other things you can think of that needs to be done that aren't in tickets already, let me know and we can create tickets for them + distribute among us this week

richardyjtian:
I made some changes with the merge commit including adding the ability for users to change their email registered on firebase. @andrewzulaybar, I modified some of your code in UserService.ts - could you re-review this file when you have time? I also moved RegistrationService.ts into UserService.ts @ericrkuo for when you connect registration to the backend.

src/services/UserService.ts
```diff
@@ -93,21 +94,51 @@ class UserService {
...
+      return axios
+        .patch(url, user)
+        .then(({ data }: AxiosResponse<UpdateUserResponse>) => {
+          if (isError<UpdateUserResponse>(data)) {
+            return Promise.reject(data);
+          }
+          return Promise.resolve();
+        })
+        .catch((err: Error) => Promise.reject(err));
+    });
```
ericrkuo: Yeah the very tricky part about this is when data starts to mismatch data between firebase and our DB.

For example, if firebase succeeds to update the email, but BE fails to update the email in the BE for some reason, we now have two mismatched emails. IMO I think we can get rid of email completely in the BE table since all we really need to connect firebase and our DB is the firebase id. Let's tackle this in the future, maybe we can create a ticket for this just so we don't forget?

src/services/UserService.ts
```diff
@@ -93,21 +94,51 @@ class UserService {
...
-  public async updateUser(user: Partial<User>): Promise<void> {
-    const url = Routes.USERS.UPDATE(this.userId);
-    return axios
-      .patch(url, user)
-      .then(({ data }: AxiosResponse<UpdateUserResponse>) => {
-        if (isError<UpdateUserResponse>(data)) {
-          return Promise.reject(data);
-        }
-        return Promise.resolve();
+  public async registerUser(firebase: Firebase, data: UserInputData): Promise<void> {
```
ericrkuo: Sorry not relevant to this code I highlighted at, but could we put a TODO here to remove this line https://github.com/ubclaunchpad/PromoPal-frontend/blob/2f49e3e2b2d0f30b5a70b387ae3c8adcde8a5b65/src/services/UserService.ts#L21

src/services/FirebaseService.ts
```diff
@@ -0,0 +1,100 @@
...
+  doSignInWithEmailAndPassword(
+    email: string,
+    password: string,
+    staySignedIn: boolean
+  ): Promise<firebase.auth.UserCredential> {
+    if (staySignedIn) {
+      this.auth.setPersistence(firebase.auth.Auth.Persistence.LOCAL);
+    } else {
+      this.auth.setPersistence(firebase.auth.Auth.Persistence.SESSION);
+    }
```
ericrkuo: Do you mind explaining how this will work?

src/services/FirebaseService.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  doReauthenticateUser(password: string): Promise<firebase.auth.UserCredential> {
+    const authUser = this.auth.currentUser;
+    // TODO: Redirect to 404 page if a user or user email does not exist
+    if (authUser === null) {
+      return Promise.reject(new Error('No user is logged in.'));
+    }
+    if (authUser.email === null) {
+      return Promise.reject(new Error('User email does not exist.'));
+    }
```
ericrkuo: Just to get my understanding better, in what situations will `authUser` and `authUser.email` be null?

src/components/account/LoginCard.tsx
ericrkuo: @yunishimura0716 I will be working on the sign in / sign up auth flow in new PR connecting it to the BE and following your auth flow doc!

src/contexts/AuthUserContext/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return <AuthUserContext.Provider value={authUser}>{children}</AuthUserContext.Provider>;
+}
+
+/**
+ * Use this function to access the context object.
+ */
+export function useAuthUser(): AuthUser | null {
+  const context = useContext(AuthUserContext);
+  return context;
+}
```
ericrkuo:
```suggestion
export function useAuthUser(): AuthUser | null {
  return useContext(AuthUserContext);
}
```

src/components/navigation/NavigationBar.tsx
```diff
@@ -29,7 +31,10 @@ export default function NavigationBar(): ReactElement {
   const location: { pathname: string } = useLocation();
 
   const [current, setCurrent] = useState<Pages>(Pages.Home);
+  const firebase = useFirebase();
+  const authUser = useAuthUser();
 
+  // TODO: isActive is not highlighting the active page when the browser back button is pressed
   const isActive = (key: Pages): CSSProperties => ({
     fontWeight: current === key ? 'bold' : 'normal',
```
ericrkuo, line 41: Nice catch, could we create a bug ticket and link it?

ericrkuo (approved):

src/services/UserService.ts
richardyjtian: Yes only tracking the email in FB sounds good. We can also approach it like in the user registration design, where if it succeeds in FB, we make a request to BE, and if it fails in BE, we abandon in FB.

src/services/FirebaseService.ts
richardyjtian: So session means the user will be logged out after the window is closed, and local means the user will stay signed in when the window is closed and re-opened.

src/services/FirebaseService.ts
richardyjtian: This should not ever happen in this function because it is called in the account details page where the user will be logged in and requesting to change their password or email. This is here more or less to address typescript errors and unauthorized users that somehow access the account page.

src/components/navigation/NavigationBar.tsx
richardyjtian: Yup, I've created https://promopal.atlassian.net/browse/PP-101 for it

ericrkuo:
@richardyjtian sorry I must have broken the build, didn't realize we need to include it in the dependency array https://github.com/ubclaunchpad/PromoPal-frontend/runs/2265756620

richardyjtian:
> @richardyjtian sorry I must have broken the build, didn't realize we need to include it in the dependency array https://github.com/ubclaunchpad/PromoPal-frontend/runs/2265756620

No problem, I'll fix it before I merge

ericrkuo:
> > @richardyjtian sorry I must have broken the build, didn't realize we need to include it in the dependency array https://github.com/ubclaunchpad/PromoPal-frontend/runs/2265756620
> 
> No problem, I'll fix it before I merge

Ah okay, thanks Richard! Would you mind tagging Andrew and I for re-review once more when you're ready since you'll be making some crucial changes?

src/services/PromotionService.ts
```diff
@@ -118,17 +119,18 @@ export async function queryPromotions(filters: FilterOptions, sort?: Sort): Prom
...
-    if (sort === Sort.Distance) {
-      const {
-        coords: { latitude, longitude },
-      } = await LocationService.GeolocationPosition.getCurrentLocation();
-      queryParams.lat = `${latitude}`;
-      queryParams.lon = `${longitude}`;
-    }
-    promotionQueryDTO.push(queryParams);
-  }
+  // todo: commented out until BE PR for sort is in
```
andrewzulaybar: We can uncomment this now since https://github.com/ubclaunchpad/PromoPal-backend/pull/127 should be ready to go!

richardyjtian: I left this commented out actually, as there seemed to be an error in push() and I didn't want to mess things up haha

src/services/UserService.ts
ericrkuo: No longer needs this because we are not storing email in our BE!

src/services/PromotionService.ts
ericrkuo: Good point, we should merge the BE PR first before uncommenting. 

The reason why we don't need push is that I changed it from type `Record<string, string>` to type `PromotionQueryDTO`, so now we should be able to do
```
promotionQueryDTO.sort = sort
promotionQueryDTO.lat = lon
promotionQueryDTO.lon = lon
```

src/components/upload-promotion/UploadPromotionForm.tsx
```diff
@@ -46,10 +46,16 @@ export default function UploadPromotionForm(): ReactElement {
...
   const history = useHistory();
+  const authUser = useAuthUser();
 
   const [datesEffective, setDatesEffective] = useState<PromotionDates>({ start: '', end: '' });
   const [times, setTimes] = useState<PromotionTimes>({});
 
+  // TODO: https://promopal.atlassian.net/browse/PP-80
+  if (!authUser) {
+    return <p>Error: No user is logged in.</p>;
+  }
```
ericrkuo, line 58: We're guaranteed that auth User is defined right, since the only way to get into this route is for auth user to be defined, otherwise it takes us to login form right

richardyjtian: Right, these if (!authUser) checks are only to see if a user illegally accesses the page

PR #40, opened 2021-03-31
PP-85 Implement deleting images

ericrkuo:
https://promopal.atlassian.net/browse/PP-85
BE PR: https://github.com/ubclaunchpad/PromoPal-backend/pull/131

### DIFF
--- a/src/services/AmazonS3Service.ts
+++ b/src/services/AmazonS3Service.ts
@@ -61,11 +61,15 @@ class AmazonS3Service {
           method: 'PUT',
           body: blobData,
         });
-        // Uploaded photo URL will be https://promopal.s3-us-west-1.amazonaws.com/`promotionId`.`imageType`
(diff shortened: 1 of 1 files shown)


src/services/AmazonS3Service.ts
```diff
@@ -61,12 +61,16 @@ class AmazonS3Service {
...
       })
       .catch((err: AxiosError) => {
         return Promise.reject(err);
       });
   }
+
+  getImageUrl(promotionId: string): string {
+    return `https://promopal.s3-us-west-1.amazonaws.com/${promotionId}`
+  }
 }
```
ericrkuo, line 74: Talked with @richardyjtian and modified the lambda function to make the `key` of the `object` just the `promotionId`. This is to support changes in the BE where do not know the image type.

@engineereng tagging you here because you're working in similar areas!

richardyjtian (approved):

PR #41, opened 2021-04-04
User Auth Workflow V2

ericrkuo:
@andrewzulaybar @richardyjtian Feel free to push changes directly to this branch or create PRs to merge into this branch. 

I put most of the TODOs we talked about in the code. Although working in parallel may be difficult, let's keep each other updated about any new changes pushed. Most of our tickets should hopefully be able to be worked on without too much merge conflicts 🙏

### DIFF
--- a/src/contexts/AuthUserContext/index.tsx
+++ b/src/contexts/AuthUserContext/index.tsx
@@ -29,7 +29,7 @@ export function AuthUserProvider({
       if (authUser) {
         axios.interceptors.request.use(
           async (request) => {
-            request.headers.authorization = await authUser.getIdToken(true);
+            request.headers.authorization = await authUser.getIdToken();
             return request;
           },
           (error) => {
--- a/src/contexts/AuthUserContext/index.tsx
+++ b/src/contexts/AuthUserContext/index.tsx
@@ -1,4 +1,5 @@
 import { User as AuthUser } from '@firebase/auth-types';
+import axios from 'axios';
 import React, { createContext, ReactElement, useContext, useEffect, useState } from 'react';
 
 import { useFirebase } from '../../contexts/FirebaseContext';
@@ -25,6 +26,17 @@ export function AuthUserProvider({
   useEffect(() => {
     firebase.getAuth().onAuthStateChanged((authUser: AuthUser | null) => {
       authUser ? setAuthUser(authUser) : setAuthUser(null);
+      if (authUser) {
+        axios.interceptors.request.use(
+          async (request) => {
+            request.headers.authorization = await authUser.getIdToken(true);
+            return request;
+          },
+          (error) => {
+            return Promise.reject(error);
+          }
+        );
+      }
     });
   });
 
--- a/src/screens/MyAccount.tsx
+++ b/src/screens/MyAccount.tsx
@@ -33,6 +33,8 @@ export default function MyAccount(): ReactElement {
    * On initial render, gets the details of the currently logged in user.
    */
   useEffect(() => {
+    // todo: If you are logged in and refresh the page, UserService.userId is no longer remembered. We need
+    //  to have AuthUserContext somehow remember the id of the user even after refreshing
     UserService.getUser()
       .then((user: User) => setUser(user))
       .catch(() => setUser(defaultUser));
--- a/src/screens/MyAccount.tsx
+++ b/src/screens/MyAccount.tsx
@@ -4,6 +4,7 @@ import React, { CSSProperties, ReactElement, useEffect, useState } from 'react';
 import AccountDetails from '../components/account/AccountDetails';
 import AccountPhoto from '../components/account/AccountPhoto';
 import ChangePassword from '../components/account/ChangePassword';
+import { useFirebase } from '../contexts/FirebaseContext';
 import UserService from '../services/UserService';
 import { User } from '../types/user';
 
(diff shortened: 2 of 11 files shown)


src/services/FirebaseService.ts
```diff
@@ -36,15 +37,15 @@ class Firebase {
     return this.auth.createUserWithEmailAndPassword(email, password);
   }
 
-  doSignInWithEmailAndPassword(
+  async doSignInWithEmailAndPassword(
```
richardyjtian, line 40: What is the benefit of making this async? I think we would still need to wait for sign in to succeed before proceeding with anything else right?

richardyjtian (approved):

src/services/FirebaseService.ts
ericrkuo: Ohya, making this async was primarily for the calls `this.auth.setPersistence(...)`. I added `await` because `setPersistance` returns a promise, otherwise, I think we could potentially be returning the call before setPersistence finishes

src/contexts/AuthUserContext/index.tsx
```diff
@@ -25,6 +26,17 @@ export function AuthUserProvider({
...
+        axios.interceptors.request.use(
+          async (request) => {
+            request.headers.authorization = await authUser.getIdToken(true);
+            return request;
+          },
+          (error) => {
+            return Promise.reject(error);
+          }
+        );
+      }
```
ericrkuo: I figured out a way for us to move the interceptor out of UserService. This `onAuthStateChange` gets called whenever the user logins, so then we will update the interceptor with the new idToken. Hypothetically, if the idToken expires, we will update the interceptor with the new idToken, we just need to make sure `onAuthStateChange` gets called. I think we can test out token expiration in a separate ticket, but this should do for now.

src/services/UserService.ts
```diff
@@ -96,31 +95,61 @@ class UserService {
...
-      })
-      .catch((error: Error) => {
-        return Promise.reject(error);
-      });
+  public async registerUser(firebaseService: FirebaseService, data: UserInputData): Promise<void> {
+    try {
+      const userCredential: firebase.auth.UserCredential = await firebaseService.doCreateUserWithEmailAndPassword(
+        data.email,
+        data.password
+      );
```
ericrkuo, line 119: If I uncomment the whole function except this chunk, it will direct me to the user details account page. My best guess is this is because firebaseService is used in a provider, I think once we remove that provider it should go away

src/screens/MyAccount.tsx
```diff
@@ -33,6 +33,8 @@ export default function MyAccount(): ReactElement {
    * On initial render, gets the details of the currently logged in user.
    */
   useEffect(() => {
+    // todo: If you are logged in and refresh the page, UserService.userId is no longer remembered. We need
+    //  to have AuthUserContext somehow remember the id of the user even after refreshing
```
ericrkuo, line 39: I added this new todo. Currently, if I sign in and select the option to stay sign in, this is equivalent to configuring firebase auth as `this.auth.setPersistence(firebase.auth.Auth.Persistence.LOCAL);` because we are persisting the auth locally, so even if I close the tab it will be remembered if I open in another tab.

This is great, but not useful if we cannot also remember the user information such as the `id` in local storage too.

src/services/FirebaseService.ts
richardyjtian: Ah, makes sense

src/contexts/AuthUserContext/index.tsx
richardyjtian: Hmm but onAuthStateChanged() will only be called when the user logs in and logs out, so the idToken would stay in the header even after expiry. I think it's better to call getIdToken() every time a request is made to the backend to ensure the token isn't expired.

src/screens/MyAccount.tsx
richardyjtian: Good point. I think we can make a request to the backend in AuthUserContext when it first runs to get the user id if the user is already signed in.

richardyjtian (changes requested):
@ericrkuo I just had one concern regarding the idToken in AuthUserContext, but besides that, I think we can merge this in

src/contexts/AuthUserContext/index.tsx
ericrkuo: Oh I see, kk thanks that makes sense. This post is helpful: https://stackoverflow.com/questions/49656489/is-the-firebase-access-token-refreshed-automatically/49697718#49697718. 

https://github.com/ubclaunchpad/PromoPal-frontend/pull/41/commits/1d32928516651ede08abcbedfd67af4c1ec14237
I've made it now so that each request to BE that requires an idToken will call firebase getIdToken(). If the token has expired, getIdToken will refresh the token and return a new one. This will trigger `onAuthStateChange` which will then update the interceptor with the newest idToken. That way as developers, we don't need to be responsible for always adding the authorization header, we just need to know to call the token each time we make a request.

ericrkuo: https://github.com/ubclaunchpad/PromoPal-frontend/pull/41/commits/de81cc1976e810896fdab72baac711be643b4d02
Also changed here, `authUser.getIdToken(true);` we don't need to specify true.

richardyjtian (approved):

PR #45, opened 2021-04-06
fix: Fix get promotions page and add loading icon

ericrkuo:
What's in this PR:
- If we called GET /promotions with the userId, our request would fail if we added any additional query parameters, for e.g. if we add additional filters like cuisine type or day of week. 
   - I think there was something strange going on with how we were constructing the URL
   - I did some refactoring and used axios's ability to specify request params
- Added loading icon when we add filters, use full text search, when refresh page and promotions are loading, etc

GetPromoSpinner

### DIFF
--- a/src/components/PromotionList.tsx
+++ b/src/components/PromotionList.tsx
@@ -1,5 +1,5 @@
 import { Place } from '@googlemaps/google-maps-services-js';
-import { Pagination } from 'antd';
+import { Pagination, Space, Spin } from 'antd';
 import React, {
   CSSProperties,
   ReactElement,
@@ -65,6 +65,11 @@ export default function PromotionList(props: Props): ReactElement {
     height: props.dimensions.height,
     width: props.dimensions.width,
     ...styles.container,
+  };
+
+  const loadingStyle: CSSProperties = {
+    top: '50%',
+    position: 'relative',
   };
 
   /**
@@ -168,11 +173,9 @@ export default function PromotionList(props: Props): ReactElement {
   useEffect(() => {
     if (promotionsState.searchQuery) {
       promotionsDispatch({ type: PromotionsDispatch.DATA_LOADING });
-      PromotionService.getPromotions([
-        {
-          searchQuery: promotionsState.searchQuery,
-        },
-      ]).then((promotions) => {
+      PromotionService.getPromotions({
+        searchQuery: promotionsState.searchQuery,
+      }).then((promotions) => {
         promotionsDispatch({ type: PromotionsDispatch.DATA_SUCCESS });
         setPromotions(promotions);
       });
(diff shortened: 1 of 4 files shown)


src/components/PromotionList.tsx
```diff
@@ -181,33 +184,42 @@ export default function PromotionList(props: Props): ReactElement {
...
-          name={promotion.name}
-          placeId={promotion.restaurant.id}
-          // TODO: https://promopal.atlassian.net/browse/PP-96
-          restaurantName=""
-          schedules={promotion.schedules}
-          onSaveButtonClick={() => onSaveButtonClick(promotion.id)}
-          onCardClick={() => onClickHandler(promotion.restaurant)}
+      {promotionsState.isLoading && !promotionsState.hasError && (
+        <Space style={loadingStyle} size="middle">
+          <Spin size="large" />
```
andrewzulaybar: Thanks for adding this!!! I'm wondering if we can use the same spinner that we're using right now in App.less, I think we just have to pass in an indicator icon

src/services/PromotionService.ts
```diff
@@ -25,21 +25,17 @@ import GooglePlacesService from './GooglePlacesService';
...
-  }
+  const endpoint = Routes.PROMOTIONS.GET;
 
   return axios
-    .get(endpoint)
+    .get(endpoint, {
+      params: {
+        ...query,
+        userId: UserService.userId,
+      },
```
andrewzulaybar, line 35: AH YES OF COURSE this is a lot better 😅

src/services/PromotionService.ts
```diff
@@ -25,21 +25,17 @@ import GooglePlacesService from './GooglePlacesService';
  *
```
andrewzulaybar, line 24: We can remove the unused import above!

src/services/PromotionService.ts
```diff
@@ -25,21 +25,17 @@ import GooglePlacesService from './GooglePlacesService';
...
  * @param query [optional] - An array of objects with key-value pairs for the query parameters
  */
-export async function getPromotions(query?: GetPromotionDTO[]): Promise<Promotion[]> {
+export async function getPromotions(query?: GetPromotionDTO): Promise<Promotion[]> {
   const userId = UserService.userId;
```
andrewzulaybar: We can remove this too 😛

andrewzulaybar (approved):
Thanks for doing this and cleaning things up -- I like it a lot better like this! Just left a few comments, but should be good to go after addressing them

andrewzulaybar:
Oh also let me know if you find other things that seem off! Especially styling I can help out with that :)

src/components/PromotionList.tsx
ericrkuo: Ohhh cool, I'll take a look at it!

src/services/PromotionService.ts
ericrkuo: Ohya good point! We'll need to add it back once we merge in the BE sort PR

src/components/PromotionList.tsx
ericrkuo: Good call! This icon looks prettier now!

ericrkuo: GetPromoSpinner2
This is what it looks like now!

ericrkuo:
> Oh also let me know if you find other things that seem off! Especially styling I can help out with that :)

@andrewzulaybar for sure! I'll be continuing to look for any bugs or areas we can improve

PR #46, opened 2021-04-06
fix: Make website open in new tab and can search on empty promotions

ericrkuo:
If I enter something in the search bar like "fry", I cannot "undo" my search to see all the promotions again. I thought it would make sense as a user, if I delete everything in my search bar and press enter, then it'll just get me all the promotions again


ericrkuo:
@andrewzulaybar something I think would be really nice in improving the responsiveness of our website is to make our restaurant cards have loader icons, and also when we click on the promotions or photos part of the restaurant card, for those to have loader icons as well. 

After adding in the loader icons for promotions, it makes our application look much better. Is this something we should prioritize?

andrewzulaybar (approved):

PR #47, opened 2021-04-06
Implement upvote/downvote functionality

andrewzulaybar:
**Description:**
- Implemented upvote/downvote buttons in the promotion card
- Converted the PromotionService file to an exported class

**Notes:**
- There weren't really many icons in Ant Design similar to what design had for the mockup, so I just used this one for being able to get this out quickly!

**Demo:**
https://www.loom.com/share/9207015f360644b19208d0870bcfd5c8


src/screens/MyAccount.tsx
```diff
@@ -18,9 +19,9 @@ const styles: { [identifier: string]: CSSProperties } = {
 
 export default function MyAccount(): ReactElement {
   const authUser = useAuthUser();
-  // TODO: https://promopal.atlassian.net/browse/PP-80
+
   if (!authUser) {
-    return <p>Error: No user is logged in.</p>;
+    return <Redirect to="/account" />;
```
richardyjtian: Ah this is a much better idea! This would essentially refresh the page right?

src/screens/MyPromotions.tsx
```diff
@@ -202,20 +263,19 @@ export default function MyPromotions(): ReactElement {
...
               onSaveButtonClick={() => onSaveButtonClick(promotion.id)}
+              onUpvoteClick={() => onUpvoteClick(promotion.id)}
             />
           </Col>
         ))}
       </Row>
     );
   };
 
-  // TODO: https://promopal.atlassian.net/browse/PP-80
```
richardyjtian, line 214: Do we maybe want to do what we do in MyAccount with the `return <Redirect to="/account" />;` ? Although it's no big deal since this should be happening.

richardyjtian (approved):

src/screens/MyAccount.tsx
andrewzulaybar: I think so yeah -- technically this is one of those weird cases because we shouldn't even be on this page if authUser is not defined (since you guarded against it in Router.tsx)

src/contexts/PromotionsListContext/index.tsx
```diff
@@ -53,7 +53,7 @@ export function PromotionsListProvider({
    */
   useEffect(() => {
     dispatch({ type: DispatchAction.DATA_LOADING });
-    getPromotions(authUser?.user?.id)
+    PromotionService.getPromotions(authUser?.user.id || '')
```
ericrkuo:
```suggestion
    PromotionService.getPromotions(authUser?.user?.id)
```

I don't think we should make it empty string, I would recommend to keep it undefined. Otherwise it'll complain like this:
image

src/components/PromotionList.tsx
```diff
@@ -105,28 +105,91 @@ export default function PromotionList(props: Props): ReactElement {
...
         if (promotion.isSavedByUser) {
           promotion.isSavedByUser = false;
-          return UserService.unsavePromotion(authUser.user.id, promotionId)
+          return UserService.unsavePromotion(authUser?.user.id || '', promotionId)
             .then(() => setPromotions(promos))
             .catch(() => null);
         }
         promotion.isSavedByUser = true;
-        return UserService.savePromotion(authUser.user.id, promotionId)
+        return UserService.savePromotion(authUser?.user.id || '', promotionId)
```
ericrkuo: I would keep what we had above or even modify it to this: Otherwise we're going to be constructing an invalid url
```
      if (!authUser?.user?.id) {
        return Promise.reject(new Error('No user is logged in.'));
      }
```

src/components/PromotionList.tsx
```diff
@@ -105,28 +105,91 @@ export default function PromotionList(props: Props): ReactElement {
...
+          promotion.votes = promotion.votes - 1;
+          promotion.voteState = VoteState.DOWN;
+          PromotionService.downvotePromotion(promotionId, authUser?.user.id || '')
+            .then(() => setPromotions(promos))
+            .catch(() => null);
+        }
+      }
+    },
+    [authUser, promotions, setPromotions]
+  );
```
ericrkuo: What do you think about this?

Less duplication and same comment about defaulting userId to empty stirng. We can also try this for upvote

```suggestion
  const onDownvoteClick = useCallback(
    (promotionId: string) => {
      if (!authUser?.user?.id) {
        return Promise.reject(new Error('No user is logged in.'));
        // or do what richard suggested to navigate to login
      }
      const promos = [...promotions];
      const promotion = promos.find(({ id }) => id === promotionId);
      if (promotion) {
        if (promotion.voteState === VoteState.DOWN) {
          promotion.votes = promotion.votes + 1;
          promotion.voteState = VoteState.INIT;
        } else {
          if (promotion.voteState === VoteState.UP) {
            // Remove upvote
            promotion.votes = promotion.votes - 1;
          }
          promotion.votes = promotion.votes - 1;
          promotion.voteState = VoteState.DOWN;
        }

        PromotionService.downvotePromotion(promotionId, authUser.user.id)
          .then(() => setPromotions(promos))
          .catch(() => null);
      }
    },
    [authUser, promotions, setPromotions]
  );
```

src/screens/MyPromotions.tsx
```diff
@@ -150,25 +150,86 @@ export default function MyPromotions(): ReactElement {
...
+          promotion.votes = promotion.votes - 1;
+          promotion.voteState = VoteState.DOWN;
+          PromotionService.downvotePromotion(promotionId, authUser?.user.id || '')
+            .then(() => setUploadedPromotions(promos))
+            .catch(() => null);
+        }
+      }
+    },
+    [authUser, uploadedPromotions, setUploadedPromotions]
+  );
```
ericrkuo: Same comment for these ones too about empty string for userId and remove duplication

src/contexts/PromotionsListContext/index.tsx
andrewzulaybar: Oh yes good catch thank you!!! Will fix

src/components/PromotionList.tsx
```diff
@@ -118,15 +121,97 @@ export default function PromotionList(props: Props): ReactElement {
...
+          promotion.votes = initialVotes - 1;
+          promotion.voteState = VoteState.DOWN;
+        }
+        return PromotionService.downvotePromotion(promotionId, authUser.user.id)
+          .then(() => setPromotions(promos))
+          .catch(() => {
+            promotion.votes = initialVotes;
+            promotion.voteState = initialVoteState;
+            message.error('An error occurred! Please try again later.', 3);
+          });
```
ericrkuo, line 175: excellent 👍

src/services/PromotionService.ts
```diff
@@ -18,117 +19,162 @@ import { isError } from '../utils/api';
...
-    })
-    .then(({ data }: AxiosResponse<GetPromotionsResponse>) => {
-      if (isError<GetPromotionsResponse>(data)) {
-        return Promise.reject(data);
-      }
-      return Promise.resolve(data);
-    })
-    .catch((err: Error) => Promise.reject(err));
-}
+class PromotionService {
```
ericrkuo, line 22: Just to be extra sure, you did not change any of the implementations right? Just moved them into a class right

ericrkuo (approved):

src/services/PromotionService.ts
andrewzulaybar: Yeah I didn't -- just moved it into a class + added the upvote/downvote ones!

PR #48, opened 2021-04-09
feat: Make restaurant card have loader icon

ericrkuo:


ericrkuo:
@andrewzulaybar Everything seems to work, but the problem is the styling because the height of the card is determined by its content.

If you have time, I was also thinking that even if the user clicks on a card with the same restaurant, we should still dismiss and open the card again. I feel like its weird that some cards are clickable and some aren't, and feel like its easier for me to be able to click on any card regardless if it's the same restaurant. What do you think?

andrewzulaybar:
@ericrkuo Fixed the height if you want to take look! Also yeah I can look into that tonight if I have time, I remember doing that before but I think adding something might have overridden it :(

ericrkuo:
> @ericrkuo Fixed the height if you want to take look! Also yeah I can look into that tonight if I have time, I remember doing that before but I think adding something might have overridden it :( 

Oh cool!!! thanks for fixing that. I think I know where we need to change to fix what I suggested, I can do it tonight :)

ericrkuo:
> If you have time, I was also thinking that even if the user clicks on a card with the same restaurant, we should still dismiss and open the card again. I feel like its weird that some cards are clickable and some aren't, and feel like its easier for me to be able to click on any card regardless if it's the same restaurant. What do you think?

I think I'll do this in another ticket

andrewzulaybar (approved):

PR #49, opened 2021-04-09
feat: Connect user auth to BE

ericrkuo:
Will test out the uploaded promotions soon and push the changes here, if any

### DIFF
--- a/src/Router.tsx
+++ b/src/Router.tsx
@@ -17,7 +17,7 @@ export default function Router(): ReactElement {
         <NavigationBar />
         <Switch>
           <Route path="/account">{authUser ? <MyAccount /> : <Login />}</Route>
-          <Route path="/promotion">{authUser ? <MyPromotions /> : <Login />}</Route>
+          <Route path="/mypromotions">{authUser ? <MyPromotions /> : <Login />}</Route>
           <Route path="/promotion/upload">{authUser ? <UploadPromotion /> : <Login />}</Route>
           <Route path="/">
             <Home />
--- a/src/components/PromotionList.tsx
+++ b/src/components/PromotionList.tsx
@@ -166,7 +166,7 @@ export default function PromotionList(props: Props): ReactElement {
     PromotionService.queryPromotions(
       promotionsState.filter,
       promotionsState.sort,
-      authUser ? authUser.user.id : undefined
+      authUser?.user?.id
     )
       .then((promotions: Promotion[]) => {
         promotionsDispatch({ type: PromotionsDispatch.DATA_SUCCESS });
--- a/src/components/account/LoginCard.tsx
+++ b/src/components/account/LoginCard.tsx
@@ -1,5 +1,5 @@
 import { Button, Checkbox, Form, Input } from 'antd';
-import React, { CSSProperties, ReactElement } from 'react';
+import React, { CSSProperties, ReactElement, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 
 import UserService from '../../services/UserService';
@@ -33,21 +33,26 @@ interface Props {
 
 export default function LoginCard(props: Props): ReactElement {
   const history = useHistory();
+  const [isDisabled, setDisabled] = useState(false);
 
   const onFinish = (data: { email: string; password: string; staySignedIn: boolean }): void => {
+    setDisabled(true);
     UserService.loginUser(data)
       .then(() => {
         history.push('/');
+        setDisabled(false);
       })
       .catch((err: Error) => {
         // TODO: https://promopal.atlassian.net/browse/PP-80
         alert(err.message);
+        setDisabled(false);
       });
   };
 
   const onFinishFailed = (): void => {
     // TODO: https://promopal.atlassian.net/browse/PP-80
     alert('Please submit the form after filling out all fields.');
+    setDisabled(false);
   };
 
   return (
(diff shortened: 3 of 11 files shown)


src/contexts/AuthUserContext/index.tsx
```diff
@@ -30,17 +30,17 @@ export function AuthUserProvider({
...
       if (firebaseUser) {
+        axios.interceptors.request.use(
+          async (request) => {
+            request.headers.authorization = await firebaseUser.getIdToken();
+            return request;
+          },
+          (err: Error) => {
+            return Promise.reject(err);
+          }
+        );
```
ericrkuo, line 41: The axios interceptor can go first before `UserService.getUser`. This is because say we configured to remain logged in.

I visit the site the next day, my idToken is no longer valid. The first time I go into the site, it'll call GET /users/:id, but then the axios interceptor is broken because there is no idToken attached to the request headers.

Moving interceptor before GEt /users/:id solves this problem

src/services/UserService.ts
```diff
@@ -145,32 +147,43 @@ class UserService {
    * @param data - Data to update the user with
    */
   public async updateUser(authUser: AuthUser, data: UserInput): Promise<void> {
-    FirebaseService.getAuth().currentUser?.getIdToken();
-    return FirebaseService.doEmailUpdate(data.password, data.email).then(() => {
+    try {
+      FirebaseService.getAuth().currentUser?.getIdToken();
+      if (authUser.firebaseUser.email !== data.email) {
+        await FirebaseService.doEmailUpdate(data.password, data.email);
+      }
```
ericrkuo: Not updating email if its the same

src/services/UserService.ts
```diff
@@ -105,11 +105,13 @@ class UserService {
   public async registerUser(data: UserInput): Promise<void> {
     try {
       const url = Routes.USERS.POST;
-      await axios.post(url, data);
+      await axios.post(url, { ...data, confirmPassword: undefined });
+      // after successful creation of user, sign in to trigger AuthUserContext and update state
+      await FirebaseService.doSignInWithEmailAndPassword(data.email, data.password, false);
```
ericrkuo, line 110: This will trigger onIdTokenChanged after successful creation of the user

src/components/PromotionList.tsx
```diff
@@ -166,7 +166,7 @@ export default function PromotionList(props: Props): ReactElement {
     PromotionService.queryPromotions(
       promotionsState.filter,
       promotionsState.sort,
-      authUser ? authUser.user.id : undefined
+      authUser?.user?.id
```
richardyjtian, line 169: Ah, I didn't know you could do this. This looks much cleaner

src/components/account/LoginCard.tsx
```diff
@@ -33,21 +33,26 @@ interface Props {
 
 export default function LoginCard(props: Props): ReactElement {
   const history = useHistory();
+  const [isDisabled, setDisabled] = useState(false);
```
richardyjtian, line 36: This is to process the login request and not allow the user to press login multiple times right?

src/contexts/AuthUserContext/index.tsx
richardyjtian: Makes sense! Good catch

src/services/UserService.ts
richardyjtian: This works, but the reason I didn't check if the email is the same is because I wanted to call the doReauthenticateUser() function (called by doEmailUpdate()) before allowing the user to update the other fields.

richardyjtian (approved):

src/Router.tsx
```diff
@@ -17,7 +17,7 @@ export default function Router(): ReactElement {
         <NavigationBar />
         <Switch>
           <Route path="/account">{authUser ? <MyAccount /> : <Login />}</Route>
-          <Route path="/promotion">{authUser ? <MyPromotions /> : <Login />}</Route>
+          <Route path="/mypromotions">{authUser ? <MyPromotions /> : <Login />}</Route>
           <Route path="/promotion/upload">{authUser ? <UploadPromotion /> : <Login />}</Route>
```
ericrkuo, line 21: For some reason, if you click promotion to promotion/upload or vice versa, it wouldn't change. I''ve separated the routes since they are actually entirely differente

src/components/PromotionList.tsx
ericrkuo: hehe yeah magic of elvis operators, I think that's what theyre called

src/components/navigation/NavigationBar.tsx
```diff
@@ -13,18 +13,21 @@ enum Pages {
...
   Account = '/account',
   UploadPromotion = '/promotion/upload',
+  MyPromotions = '/mypromotions',
 }
 
 const PathsToPages: { [path: string]: Pages } = {
   [Paths.Home]: Pages.Home,
   [Paths.Account]: Pages.Account,
   [Paths.UploadPromotion]: Pages.UploadPromotion,
+  [Paths.MyPromotions]: Pages.MyPromotions,
```
ericrkuo, line 30: Now looks like
image

src/components/account/LoginCard.tsx
ericrkuo: yeah exactly, same for register card, we don't want to allow them to spam it n times and make n requests to BE

src/services/UserService.ts
ericrkuo: Ohh good point, would this work?
```suggestion
      if (authUser.firebaseUser.email !== data.email) {
        await FirebaseService.doEmailUpdate(data.password, data.email);
      } else {
        await FirebaseService.doReauthenticateUser(data.password);
      }
```

PR #50, opened 2021-04-10
feat(image-upload): Integrate image storing with FE components

richardyjtian:
Image storage is integrated with FE and should be good to go! Just some styling changes are needed

### DIFF
--- a/src/components/ImageSelector.tsx
+++ b/src/components/ImageSelector.tsx
@@ -1,51 +1,40 @@
 import { Button, Image as Img } from 'antd';
-import React, { ReactElement, useState } from 'react';
+import React, { ReactElement } from 'react';
 
 import AmazonS3Service from '../services/AmazonS3Service';
 import { Image } from '../types/image';
 
 interface Props {
-  onChange: (imageBinary: string, imageType: string) => void;
+  image: Image;
+  setImage: (image: Image) => void;
 }
 
 export default function ImageSelector(props: Props): ReactElement {
-  const [image, setImage] = useState<Image>({ imageBinary: '', imageType: '' });
-
   return (
     <>
       {/* Displays the image selector if no image is selected */}
-      {!image.imageBinary && (
+      {!props.image.imageBinary && (
         <input
           type="file"
           accept="image/jpeg, image/png, image/gif"
           onChange={(event) => {
             AmazonS3Service.onFileChange(event)
               .then((image: Image) => {
-                setImage(image);
-                props.onChange(image.imageBinary, image.imageType);
+                props.setImage(image);
               })
               .catch((err: Error) => alert(err));
           }}
         />
       )}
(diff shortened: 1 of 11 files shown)


src/components/ImageSelector.tsx
```diff
@@ -33,7 +43,7 @@ export default function ImageSelector(): ReactElement {
           after an promotion has been successfully saved to our backend db
       */}
       {image.imageBinary && (
-        <button onClick={() => AmazonS3Service.uploadImage(image, 'promotionId123')}>
+        <button onClick={() => AmazonS3Service.uploadImage(image, 'testpromoid')}>
```
andrewzulaybar: Oh yeah we don't need this button anymore right since we're uploading the image in PromotionService.PostPromotion?

andrewzulaybar (approved):

src/components/ImageSelector.tsx
richardyjtian: oops I forgot to push my latest changes! but yes! I've removed it

ericrkuo:
Reviewing right now!

src/components/promotion/PromotionImage.tsx
```diff
@@ -19,6 +19,7 @@ export default function PromotionImage(props: Props): ReactElement {
   const defaultSrc =
     'https://d1ralsognjng37.cloudfront.net/92a7b4fb-892c-47bb-b5bb-884c89c254a2.jpeg';
 
+  // TODO: If getting the image from S3 fails, resort to a default image
```
ericrkuo, line 22: How difficult do you think this TODO would be, just curious because I think it would be beneficial. 

I think it may be as simple as this:
```
<div>{imageSrc ? <img style={styles.image} src={imageSrc} alt=""  onerror="if (this.src != 'error.jpg') this.src = 'error.jpg';" /> : <Skeleton.Image />}</div>
```

src/services/PromotionService.ts
```diff
@@ -68,15 +70,19 @@ export async function deletePromotion(id: string): Promise<void> {
...
   const url = Routes.PROMOTIONS.POST;
   return axios
     .post(url, promotionDTO)
-    .then(({ data }: AxiosResponse<PostPromotionsResponse>) => {
+    .then(async ({ data }: AxiosResponse<PostPromotionsResponse>) => {
       if (isError<PostPromotionsResponse>(data)) {
         return Promise.reject(data);
       }
+      if (image) {
+        await AmazonS3Service.uploadImage(image, data.id);
```
ericrkuo: Maybe a tiny comment so we don't forget about how if image upload fails, we still resolve

ericrkuo (approved):

src/components/promotion/PromotionImage.tsx
andrewzulaybar: Oh yes I'll add that!

